Issue #3, opened 2017-10-12
Volume data type support


Enet4:
I will use this issue to track currently supported volume data types.

| type | NiftiVolume (`get_*`) | `ToNdArray` | Tested | Issue |
|-------|--------------------------|-------------------|----------|---------|
| `Uint8` | :white_check_mark: | :white_check_mark: | :white_check_mark: | |
| `Int8` | :white_check_mark: | :x: | :x: | |
| `Uint16` | :white_check_mark: | :x: | :x: | |
| `Int16` | :white_check_mark: | :x: | :x: | |
| `Uint32` | :white_check_mark: | :x: | :x: | |
| `Int32` | :white_check_mark: | :x: | :x: | |
| `Uint64` | :white_check_mark: | :x: | :x: | |
| `Int64` | :white_check_mark: | :x: | :x: | |
| `Float32` | :white_check_mark: | :white_check_mark: | :white_check_mark: | #1 |
| `Float64` | :white_check_mark: | :x: | :x: | |
| others | :x: | :x: | :x: | |

nilgoyette:
I see that you support `u8` image but I can't get this snippet to work.

    let obj = InMemNiftiObject::from_file("a path.nii.gz")
        .expect("Can't read input file.");
    let volume_dyn = obj.into_volume().to_ndarray::<u8>().unwrap();

It fails during the compilation with error `the trait std::convert::From<f32> is not implemented for u8`. I don't even use of `f32` in this block! I check your tests and they all use `to_ndarray::<f32>`

Enet4:
Thanks for the feedback, @nilgoyette.
Perhaps the table can indeed lead to misinterpretation: while the library can read volumes of the data types above, it can currently only fetch floating point values from the volume. That is, obtaining an `Array<u8, _>` is not supported, and there is no `get_u8` method either. Since values are subjected to the linear transformation over the scale+intercept parameters, a conversion to  is not guaranteed in compile time not to lose precision, thus the constraint `From<f32>`.

This is something that can be reconsidered, of course, and the crate could sure use more tests. May I know what brings you to reading a volume as an array of `u8`? Perhaps you would like to retrieve the raw volume data without the linear transformation?

nilgoyette:
I was trying to read a binary mask (0-1 values). They are quite frequent in medical imaging.

[...]

PR #5, opened 2017-11-07
Support all POD for ndarray

nilgoyette:
Hi Enet4. my company needed to load more datatypes and I didn't want to disturb you so I added them myself. The way I did it, it first loads the header-defined datatype so there's no precision loss, then it casts to the desired type, which may be bad for the precision. However, this is not our fault because it's the user-desired type... he's supposed to know what he's loading. You can use it the same way you did before: `volume.to_ndarray::<u32>()` or any other type.

I understand that this is not idiomatic Rust. I'm a python and C++ programmer by trade so my rust is not as good as it could be. Please be as critical as you would be to your own code.

You were going, I believe, in another direction with the type management. I'm not too sure where you where going though so I coded it like I would in C++. If you actually prefer another way, just tell me and close this PR. My company will use my fork though because it's working quite well and we need it now :)

This should be tested a lot, so I tested everything... **except the weird types (complex, rgb and 128)** I only used `u8`, `i16` and `f32` in real-life tests though. 128bits would be quite easy to add.

If you actually merge this, I'll probably send a PR specific dimension (instead if dynamic) and a PR for the rgb types because we might need them.


Enet4:
Hello, and thank you very much for your contribution! I can see that some effort was put to this, and I'm mostly interested in them.

However, I might take a bit more time to pick this up than usual. This includes taking some of my spare time to review these changes properly and to reflect some more about the breaking changes introduced. Right now I'm feeling reluctant to add the constraint `T: PodTransmutable`, mostly because it's currently meant to be an implementation detail at this time. Being stuck with `AsPrim` might not be a good idea either, so maybe it's an opportunity to think this through for an ideal solution. This is one of the reasons that I didn't land 1.0.0 yet.

nilgoyette:
`PodTransmutable` could be removed, but we would be forced to code `guarded_transmute_pod_vec_permissive()` ourself, which is sad because this function rocks.

I forgot to explain what are the "27" images.  They are simply (in python) `np.arange(27).reshape((3, 3, 3))` saved with nibabel for each type. 0..27 is ok for all types so I can test everything loading and casting.

Moreover... my bad. I always run the tests with the --features flag. I thought this was testing everything. That's not actually what's happening :) I'll fix this.

Enet4:
> PodTransmutable could be removed, but we would be forced to code 
`guarded_transmute_pod_vec_permissive()` ourself, which is sad because this function rocks.

Yes, it does. And it's not just because I'm one of the maintainers of `safe_transmute`. 😅 
I don't necessarily mean not to use the function, but to devise a constraint that will allow us to use an efficient implementation while still enabling a flexible use and support for all data types, including the "trickier" ones.

> I forgot to explain what are the "27" images. They are simply (in python) np.arange(27).reshape((3, 3, 3)) saved with nibabel for each type. 0..27 is ok for all types so I can test everything loading and casting.

Good to know!

> Moreover... my bad. I always run the tests with the --features flag. I thought this was testing everything. That's not actually what's happening :) I'll fix this.

That happens to me sometimes as well. I'm glad Travis picks these up nicely. :)

src/util.rs
```diff
@@ -183,20 +187,42 @@ where
...
+    e: Endianness
+) -> Vec<T> {
+    let nb_bytes = mem::size_of::<T>();
+    if e != Endianness::system() && nb_bytes > 1 {
+        // Swap endianness by block of nb_bytes
+        let split_at = nb_bytes / 2;
+        let split_at_m1 = split_at - 1;
+        for c in a.chunks_mut(nb_bytes) {
+            let (a, b) = c.split_at_mut(split_at);
+            for i in 0..split_at {
```
Enet4: Here's something more idiomatic:

```
for (l, r) in a.iter_mut().zip(b.iter_mut().rev()) {
    mem::swap(l, r);
}
```

No indexing operator, and no need for `split_at_m1`.

src/util.rs
```diff
@@ -183,20 +187,42 @@ where
...
+                mem::swap(&mut a[i], &mut b[split_at_m1 - i]);
+            }
         }
     }
-    let mut v: Vec<f32> = guarded_transmute_pod_vec_permissive(a);
-    for e in &mut v {
-        *e = designalise_f32(*e);
```
Enet4: It feels problematic to leave this transformation out, but I'll seek to solve this before the next release.

src/util.rs
nilgoyette: Much better, thank you.

src/util.rs
nilgoyette: I wasn't sure when I removed it because I wonder if modifying the input data is "right". I mean, there's probably a reason why there's a Signaling NaN in an image. I don't know of any, but why would it be there?

A better question might be "Why did you add the call to `designalise_f32`?"

Enet4: Usually, there isn't a signalling NaN. But a file having one is not impossible. This, along with the fact that the Rust community at large has preferred not to take any chances, is what drives me to still convert all NaNs to quiet NaNs.

As the number of data types possible is pretty limited, one can build efficient and safe implementations for reading one or more values without too much boilerplate. This will also make dealing with these corner cases possible without exposing traits related to implementation details. My current concern is that I have never found an RGB or Complex NIfTI-1 volume in practice. If you happen to know more on the subject, I would be glad to talk about it on issue #3. :)

nilgoyette:
Hey @Enet4, I know you wanted to refactor your lib before a new release but can you please publish a new version with this PR? Since I'm the creator of the [only dependant crate](https://crates.io/crates/nifti/reverse_dependencies), I don't think anyone will complain.

Enet4:
@nilgoyette All right, I am not opposed to that, although I wouldn't call public dependant crates a good metric. I was just hoping to reduce the number of breaking changes from one version to another, but I admit that I'm taking a bit longer than I expected.

A new version is up. Aside from your PR, I just updated its current dependencies to use `safe-transmute` 0.6.0 and `flate2` 1.0.1.

PR #10, opened 2018-07-18
Use 'c' order instead of 'fortran' order when creating ndarray

nilgoyette:
We are currently using 'fortran' order (column major) when converting the volume to a dynamic ndarray. This is working perfectly but a majority of users probably prefers 'c' order (row major).

All tests still pass. I wasn't sure if I needed to add a new one. I decided not to because 1) it's supposed to be transparent to the user 2) I'm not sure what to test. a pointer greater than another pointer?

### DIFF
--- a/src/volume/inmem.rs
+++ b/src/volume/inmem.rs
@@ -210,8 +210,7 @@ impl InMemNiftiVolume {
         let mut data: Vec<O> = data.into_iter().map(AsPrimitive::as_).collect();
         <O as DataElement>::Transform::linear_transform_many_inline(&mut data, self.scl_slope, self.scl_inter);
 
-        Ok(Array::from_shape_vec(IxDyn(&dim).f(), data)
-            .expect("Inconsistent raw data size"))
+        Ok(Array::from_shape_vec(IxDyn(&dim), data).expect("Inconsistent raw data size"))
     }
 }
 


Enet4:
Thanks for the PR. Something still isn't quite right, though (did you enable the `ndarray_volumes` feature in your tests?). Since the use of `.f()` during the shape building process makes a distinct reinterpretation of the inner vector, one has to make sure that the volumes in the tests are also built in row major order. I'll be making a more thorough review in a moment...

nilgoyette:
I ran `cargo test` and `cargo test --features ndarray`. I wasn't aware that the right name is `ndarray_volumes`, sorry. I'll check again tomorrow at work.

Enet4:
Defining a name for this feature was also a decision I had to think about for a moment. According to the Cargo documentation on [features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section), a unique feature name provides a higher level interpretation of the functionality, thus why I went for the slightly more verbose `ndarray_volumes` instead of `ndarray`. Notwithstanding that this decision could bite someone eventually, because the feature is indeed, support for `ndarray`. It's not too late to change this too.

nilgoyette:
Hum, it seems you're right. I tested with nibabel

    >>> im = nib.load('resources/minimal.nii.gz')
    >>> im.get_data().flags
      C_CONTIGUOUS : False
      F_CONTIGUOUS : True
      ...

`array_from_file` and `array_to_file` with a default `order='F'` parameter implies that all images loaded or saved by NiBabel is on Fortran order. It can't actually be changed when using the normal methods. I got the same result with itk. I asked my friend who knows the nifti standard more than me and it turns out that **nifti requires the image to be in Fortran order** on disk. This explains why everyone does it that way!

We concluded that we need to call `.f()` when creating our output arrays if we want to have a fast writer. Otherwise, we need to call `.t()` and iterate. What's misleading is that nifty-rs and ndarray disagree on their default order, much like nibabel and and numpy. There's nothing we can do about it though! We just need to be aware of it.

Thank you for your help.

Enet4:
Well, I still believe that we can document this behaviour better on this end, so I'll be throwing a PR for additional clarity later. Even a deep look at the NIfTI-1 specifications was not as conclusive as I would have wished. Regardless, it's best for this library to output `ndarray`s in their original memory layout and let users know that they might want to adjust their pipeline accordingly. Thank you for bringing this up!

@nilgoyette On a side note, I would also like you to become an additional collaborator and crate owner of `nifti`. This is not an attempt to delegate responsibilities (it's a small and low-traffic open source project after all), but I see that you have great interest in this library and I believe that a second maintainer would provide additional confidence to those using it in a professional environment. What do you think?

nilgoyette:
@Enet4 Yeah, sure, I'm already 'Watching' this lib anyway and, as you wrote, there's no much job :) Thanks for the offer. I'll probably contribute our in-house nifti writer when we think it's ready.

PR #11, opened 2018-07-22
Improve documentation

Enet4:
- Mention `ndarray_volumes` feature in lib.rs
- Explain `LinearTransform` a bit more
- Expand `InMemNiftiVolume` docs on `ndarray` and array memory order
- Significantly expand documentation in `volume::ndarray` module
- rustfmt ndarray.rs

### DIFF
--- a/src/volume/ndarray.rs
+++ b/src/volume/ndarray.rs
@@ -1,14 +1,43 @@
-//! Interfaces and implementations specific to integration with `ndarray`
-use ndarray::{Array, Axis, Ix, IxDyn};
-use volume::NiftiVolume;
-use std::ops::{Add, Mul};
-use num_traits::AsPrimitive;
+//! Interfaces and implementations specific to integration with `ndarray`.
+//!
+//! This module introduces the trait [`IntoNdArray`], which is implemented for
+//! all NIfTI volume types and enable their mapping into an [`Array`] with a
+//! dynamic number of dimensions and an abitrary element type. The affine
+//! scaling of the values (from the `scl_slope` and `scl_inter` attributes) are
+//! also considered in this transformation.
+//! 
+//! A target [element type] needs to be provided at compile time, which is
+//! usually the data type that the user wishes to work the array with. If the
+//! source and target types do not match, each voxel is cast in a way as to
+//! avoid loss of precision.
+//!
+//! #### Note on memory order
+//! 
+//! NIfTI volumes are usually stored in disk in column major order (also called
+//! Fortran order). As such, the array resulting from this operation will also 
+//! be in this memory order, rather than the usual row major order (AKA C
+//! ordering). When accessing the array, one should consider any potential
+//! bottlenecks emerging from this ordering in their data processing pipelines.
+//! Namely, it might be faster to produce output arrays in column major order
+//! as well.
+//! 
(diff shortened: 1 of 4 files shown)


src/volume/ndarray.rs
```diff
@@ -1,14 +1,43 @@
...
+//!
+//! #### Note on memory order
+//! 
+//! NIfTI volumes are usually stored in disk in column major order (also called
+//! Fortran order). As such, the array resulting from this operation will also 
+//! be in this memory order, rather than the usual row major order (AKA C
+//! ordering). When accessing the array, one should consider any potential
+//! bottlenecks emerging from this ordering in their data processing pipelines.
+//! Namely, it might be faster to produce output arrays in column major order
+//! as well.
```
nilgoyette: This last sentence, if it's an advice to the user , should probably be accompanied with a snippet explaining how to do it. Simply ways like `arr.t().to_owned()` or `arr.reversed_axes()` simply modifies the strides. In fact, I'm not even sure how to do it properly.

Enet4: To be honest, it took me a while to get this too. :persevere: By the looks of it, we stumbled upon a use case not well covered by `ndarray` at this time. Changing an array's memory layout while keeping it logically equivalent requires more than just reversing axes, because that will only change the dimensional component (shape + strides) of the array. I think the cleanest (not necessariy efficient) implementation we can come up with is this:

```rust
fn array_to_standard<A, D>(arr: &Array<A, D>) -> Array<A, D>
where
    D: Dimension,
    A: Zero,
    A: Clone,
{
    let mut x = Array::zeros(arr.raw_dim());
    debug_assert!(x.is_standard_layout());
    for (l, r) in x.iter_mut().zip(arr) {
        *l = r.clone();
    }
    x
}
```

The `Zero` constraint can be removed with `unsafe`.

Checking memory layout is also a bit of a nuisance. While there is an easy way to check whether the array is in row major order (`is_standard_layout()`), we don't have such an easy check for column major layout:

```
fn array_is_f_order<T, D>(arr: &ArrayBase<T, D>) -> bool
where
    T: Data,
    D: Dimension,
{
    arr.as_slice_memory_order().is_some() // is contiguous
    && !arr.is_standard_layout() // but is not in C order
}
```

I get the feeling that this would be better addressed at its root, by proposing additions to ndarray. What could be added here in the mean time?

nilgoyette: I agree that there should be a way to change the memory order of an ArrayBase, in ndarray, not here. If we need it, then someone else needs it too.

What you wrote is perfect in the meantime. Thank you.

Enet4: Whelp, it turns out that there is some progress upstream: after being brought up in bluss/ndarray#441, bluss/ndarray#442 documents `to_owned`, and it proposes a better way to convert an array to standard layout:

```rust
let arr = Array::from_shape_vec((2, 2), vec![1, 2, 3, 4]).unwrap();
let owned = Array::from_shape_vec(arr.raw_dim(), arr.iter().cloned().collect()).unwrap();
assert!(owned.is_standard_layout());
assert_eq!(arr, owned);
```

It's been four months since a new version of `ndarray` was released. I wonder how we can gracefully suggest to launch a release with the latest changes. In the mean time, this PR is good to go indeed.

PR #12, opened 2018-09-04
Update ndarray dependency

nilgoyette:
Update to ndarray 0.12.

Tested with and without `--features ndarray_volumes`. I'm not sure if you wanted a "strange" version like `">=0.10.12,<0.13.0"`?


nilgoyette:
Right, good idea. I'm not really used to those "ranged" versions. Nice feature.

Enet4:
It's in. And I think it's high time we launched nifti 0.5.1! It'll be but a moment.

PR #14, opened 2018-10-09
Move endianness info to NiftiHeader

Enet4:
This PR changes `NiftiHeader` so that it contains an extra `endianness` member containing the byte order of the originally read file. When building a new header, the system's native endianness will be the default. Moreover, the respective constructors will no longer yield a `(header, endianness)` pair:

```rust
// before
pub fn from_file<P: AsRef<Path>>(path: P) -> Result<(NiftiHeader, Endianness)> { ... }
// after
pub fn from_file<P: AsRef<Path>>(path: P) -> Result<NiftiHeader> { ... }
```

This is a design decision that has kept in my mind for a while, but now I honestly believe it's worth the breaking change, for the following reasons:

- It is unconventional to see constructors such as `from_stream` and `from_file` to return a tuple rather than just `Self` or `Result<Self>`.
- Returning a tuple makes these functions more complicated to work with.
- The header's byte order is an inherent property materialized in how the contents were written to disk, and that property is lost in the current reading process (which harmonizes all attributes to native byte order).
- `NiftiHeader` does not have a native memory representation anyway. The `descrip` member was already put into a `Vec<u8>` (arrays larger than 32 are currently bad to work with), so one shouldn't be relying on this type to be a memory equivalent of the respective C-like struct. Adding an extra member is not a problem in this regard.
- And finally, those focused on using the `NiftiObject` API shouldn't notice a difference.

### DIFF
--- a/examples/niftidump/main.rs
+++ b/examples/niftidump/main.rs
@@ -9,7 +9,6 @@ fn main() {
     let mut args = env::args().skip(1);
     let filename = args.next().expect("Path to NIFTI file is required");
     let header = NiftiHeader::from_file(filename)
-        .expect("Failed to read NIFTI file")
-        .0;
+        .expect("Failed to read NIFTI file");
     println!("{:#?}", &header);
 }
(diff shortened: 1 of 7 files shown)


nilgoyette:
Good idea. LGTM.

PR #16, opened 2018-10-10
InMemNiftiVolume::from_raw_data

nilgoyette:
Usage would look like:

    let mut header = obj.header().clone();
    let mut volume = obj.into_volume();
    // modify something in header, like dim, or modify raw_data, but it's a bad idea!
    volume = InMemNiftiVolume::from_raw_data(&header, volume.to_raw_data()).unwrap();

I did what I did because I only have a single use-case to consider. I'm not too sure about:

1. receiving a header. It seemed better than asking for `dim`, `datatype`, `scl_slope`, `scl_inter` and `endianness`, but it forces the programmer to create a header. I don't think it's a problem because it's easy to create one with `default` and the `..` syntax.
2. returning a Result. I don't like it, but I don't like asking for a `datatype` either.

I added **a test** for my "false 4D image" that I didn't push because I don't know if you want to have more toy images in your lib.

### DIFF
--- a/src/volume/inmem.rs
+++ b/src/volume/inmem.rs
@@ -46,7 +46,7 @@ impl InMemNiftiVolume {
     /// declared in the header are expected to fit.
     pub fn from_raw_data(header: &NiftiHeader, raw_data: Vec<u8>) -> Result<Self> {
         if nb_bytes_for_data(header) != raw_data.len() {
-            return Err(NiftiError::IncompatibleLengthError);
+            return Err(NiftiError::IncompatibleLength);
         }
 
         let datatype: NiftiType =
--- a/src/volume/inmem.rs
+++ b/src/volume/inmem.rs
@@ -10,7 +10,7 @@ use header::NiftiHeader;
 use extension::{Extender, ExtensionSequence};
 use error::{NiftiError, Result};
 use volume::element::DataElement;
-use util::Endianness;
+use util::{Endianness, nb_bytes_for_data};
 use byteorder::{BigEndian, LittleEndian};
 use flate2::bufread::GzDecoder;
 use typedef::NiftiType;
@@ -25,12 +25,12 @@ use ndarray::{Array, Ix, IxDyn, ShapeBuilder};
 /// contain raw image data, which is converted automatically when using reading
 /// methods or [converting it to an `ndarray`] (only with the `ndarray_volumes`
 /// feature).
-/// 
+///
 /// Since NIfTI volumes are stored in disk in column major order (also called
 /// Fortran order), this data type will also retain this memory order.
-/// 
+///
 /// [converting it to an `ndarray`]: ../ndarray/index.html
-/// 
+///
 #[derive(Debug, PartialEq, Clone)]
 pub struct InMemNiftiVolume {
     dim: [u16; 8],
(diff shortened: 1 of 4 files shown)


src/volume/inmem.rs
```diff
@@ -42,6 +42,23 @@ pub struct InMemNiftiVolume {
 }
 
 impl InMemNiftiVolume {
+    /// Build a InMemNiftiVolume from a header and a buffer. The buffer length and the dimensions
```
Enet4: "Build a InmemNiftiVolume" → "Build an `InMemNiftiVolume`"

src/volume/inmem.rs
```diff
@@ -42,6 +42,23 @@ pub struct InMemNiftiVolume {
...
+    /// declared in the header are expected to fit.
+    pub fn from_raw_data(header: &NiftiHeader, raw_data: Vec<u8>) -> Result<Self> {
+        assert!(nb_bytes_for_data(header) == raw_data.len(),
```
Enet4: Oops, forgot about this line. Please make it a recoverable error by making a new error variant.

tests/object.rs
```diff
@@ -188,3 +187,31 @@ fn f32_nii_gz() {
     assert_eq!(volume.get_f32(&[5, 0, 4]).unwrap(), 0.4);
     assert_eq!(volume.get_f32(&[0, 8, 5]).unwrap(), 0.8);
 }
+
+#[cfg(feature = "ndarray_volumes")]
```
Enet4: Since this new capability isn't restricted to ndarray volumes, how about we make it happen without "ndarray_volumes"? The last part can be kept behind a `cfg!` condition:

```rust
if cfg!(feature = "ndarray_volume") {
    use ndarray::Ix3;

    let dyn_data = volume.to_ndarray::<f32>().unwrap();
    ...
}
```

nilgoyette: I don't want to use you as a debugger but it's not working.

    // Missing crate ndarray, no IntoNdArray in the root, etc.
    if cfg!(feature = "ndarray_volumes") {
        use ndarray::Ix3;
        use nifti:: IntoNdArray;
        ...

    // Compiles and runs
    #[cfg(feature = "ndarray_volumes")] {
        use ndarray::Ix3;
        ...

I can't find any proof, but it looks like it can only be used for simple condition. Have you used it with `extern crate` and/or `use` before?

Enet4: Well, it seems that you are right. Conditional expressions based on the outcome of `cfg!` do not prevent their respective blocks from being analysed, even when that outcome is known at compile time. We have to employ directives, as you suggested.

```diff
:100644 100644 49128d1 0000000 M	tests/object.rs

diff --git a/tests/object.rs b/tests/object.rs
index 49128d1..544ea14 100644
--- a/tests/object.rs
+++ b/tests/object.rs
@@ -188,11 +188,9 @@ fn f32_nii_gz() {
     assert_eq!(volume.get_f32(&[0, 8, 5]).unwrap(), 0.8);
 }
 
-#[cfg(feature = "ndarray_volumes")]
 #[test]
 fn test_false_4d() {
-    use ndarray::Ix3;
-    use nifti:: {InMemNiftiVolume, IntoNdArray};
+    use nifti::InMemNiftiVolume;
 
     let (w, h, d) = (5, 5, 5);
     let mut header = NiftiHeader {
@@ -210,8 +208,15 @@ fn test_false_4d() {
         volume = InMemNiftiVolume::from_raw_data(&header, volume.to_raw_data()).unwrap();
     }
     assert_eq!(volume.dimensionality(), 3);
-    let dyn_data = volume.to_ndarray::<f32>().unwrap();
-    assert_eq!(dyn_data.ndim(), 3);
-    let data = dyn_data.into_dimensionality::<Ix3>().unwrap();
-    assert_eq!(data.ndim(), 3); // Obvious, but it's to avoid being optimized away
+
+    #[cfg(feature = "ndarray_volumes")]
+    {
+        use ndarray::Ix3;
+        use nifti::IntoNdArray;
+
+        let dyn_data = volume.to_ndarray::<f32>().unwrap();
+        assert_eq!(dyn_data.ndim(), 3);
+        let data = dyn_data.into_dimensionality::<Ix3>().unwrap();
+        assert_eq!(data.ndim(), 3); // Obvious, but it's to avoid being optimized away
+    }
 }
```

But now I also noticed that this is in tests/object.rs. I would understand this as a unit test for the in-memory nifti volume rather than an integration test for a nifti object. Can you move it to the `test` module in "src/volume/inmem.rs"?

PR #18, opened 2018-10-16
Improve integration with code type definitions

Enet4:
- specify memory representation in `typedef::*` (u16 or u8 depending on attribute)
- make utility methods in `typedef::*` take `self` by value (cheaper than by ref)
- make volume construction methods yield an `InvalidCode` error (instead
of the less specific `InvalidFormat`)
- add `NiftiError::InvalidCode` variant
- add validating getters for these codes
- add methods for retrieving xyzt units with space-time separation

At first I considered modifying the attributes of `NiftiHeader` to contain these types immediately upon reading the header. With the repr directives, the header could still be packed without wasting memory and providing an extra layer of abstraction. However, I feel reluctant to make the entire reading process break just because of a single unknown code, which for the most part (`datatype` is an exception) are not critical. By keeping unvalidated codes, the library will give more flexibility to eventual file diagnostic tools.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -51,6 +51,11 @@ quick_error! {
         IncompatibleLength {
             description("The buffer length and the header dimensions are incompatible.")
         }
+        /// Header contains a code which is not valid for the given attribute
+        InvalidCode(typename: &'static str, code: i16) {
+            description("invalid code")
+            display("invalid code `{}` for {}", code, typename)
+        }
     }
 }
 
--- a/src/header.rs
+++ b/src/header.rs
@@ -1,13 +1,15 @@
 //! This module defines the `NiftiHeader` struct, which is used
 //! to provide important information about NIFTI-1 volumes.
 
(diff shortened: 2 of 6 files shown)


Enet4:
> Just to be sure, does calling a method on it consume the value? Or it's safe because it's a `u8`/`u16` and they are defined as `Copy` objects? Probably the later.

Yes, it is actually preferable for small `Copy` types to have consuming methods. It is perceived by the compiler as consuming a copy of the value, which is cheaper than making a reference to it (`u8` vs `&u8`). In practice, usage doesn't change, but it's more idiomatic and leaves a bit more room for machine code optimization [citation_needed]. Primitive integer types in stdlib also follow this guideline (e.g. [`i16`](https://doc.rust-lang.org/std/primitive.i16.html)), and Clippy has a [lint](https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#trivially_copy_pass_by_ref) for this.

Issue #19, opened 2018-10-16
Writing NIfTI files


Enet4:
In order to write these files, we need methods for:

- [ ] writing a `NiftiHeader` to a file (".hdr") or writer;
- [ ] writing a volume to a file (".img") or writer, with or without gzip;
- [ ] writing a full object to a file (".nii") or writer, with or without gzip.

nilgoyette:
We, at Imeka, already have the code to write a .nii or .nii.gz image and we want to contribute. It has been heavily tested (we wrote several thousands images in the last months). Of course, it's not perfect, we're not sure how to handle the scaling slope and offset, and maybe some other problems that I don't remember at the moment.

Enet4:
I would be happy to receive and review your contributions! We can do this piece by piece and discuss those concerns as they appear in the code.

nilgoyette:
I forgot to tell you, my code is totally tied to `ndarray`. We won't be able to check any of your 3 tasks above. This being said, most of it can be re-used if this is merged. As a user of this library, I mostly use `NiftiHeader` and `ndarray` objects.

    pub fn write_nifti<A, S, D>(
        path: &str,
        data: &ArrayBase<S, D>,
        reference: Option<&NiftiHeader>)
        where S: Data<Elem=A>, ...
    { ... }

Also, how to you manage temporary files in your tests? I usually use the `tempfile` crate but I'm pretty sure you don't want to add all those libs...

     Compiling winapi v0.3.6
     Compiling rand_core v0.3.0
     Compiling cfg-if v0.1.5
     Compiling rand_core v0.2.2
     Compiling remove_dir_all v0.5.1
     Compiling rand v0.5.5
     Compiling ansi_term v0.11.0
     Compiling pretty_assertions v0.5.1
     Compiling tempfile v3.0.4

Enet4:
I am not opposed to including a crate for temporary files in tests, especially because that would be a development dependency.

jbteves:
Hi, I ran across your project while trying to make my own nifti crate (which is significantly less sophisticated...) I'd like to help out with this, are there any developer docs aside from what's on the repo?

nilgoyette:
I don't think there more docs. The project is not too big so it's easy enough to jump in and modify.

We will be happy to review and merge your PRs. Is there something that you need and that nifti-rs doesn't handle?

jbteves:
[...]

PR #21, opened 2018-10-18
Write ndarray to nifti

nilgoyette:
As explained in #19, here's our code to write a `ndarray` to a nifti file. I kept it simple, it's all in the same file, you simply need to call `write_nifti` with an `Array` or an `ArrayView`, in `c` or `f` ordering. When you want to create a new image with the same header (same position, spacing, etc.), simply call the function with the optional `NiftiHeader`.

In a generic context, you will need to use the trait `TypeTool`. I'm not a fan of it, but I don't know how to avoid it, and it's not too bad, maybe it simply needs to be hidden :)

Some problems that I can see with my code:
- The `if slope != 1.0 || header.scl_inter != 0.0 {` condition is not required. Maybe we shouldn't care about some substraction and divisions...
- As I said, `TypeTool`. The name is bad, I know. But I do think it's a good trait usage.
- The ndarray `dim` and `datatype` currently win over the same fields in the header. That may or may not be what the user wants. I know that nibabel prefers the header, so it casts the data. I personally don't like automatic hidden casts, I prefer converting my ndarray myself.
- Afaik, NiBabel ignores `scl_slope` and `scl_inter` on writing. They are simply set to 1.0 or 0.0. I never ignore them. This is sadly as "automatic" as the casts...
- The last 2 points are complex. Not complex to code but to decide what should be done.

By the way, do not hesitate to be critical! I won't cry over it. It's the best way to learn.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -78,7 +78,8 @@ where
 }
 
 fn write_header<W>(writer: &mut W, header: &NiftiHeader) -> Result<()>
-    where W: WriteBytesExt
+where
+    W: WriteBytesExt,
 {
     writer.write_i32::<B>(header.sizeof_hdr)?;
     writer.write_all(&header.data_type)?;
@@ -233,7 +234,8 @@ pub mod tests {
     }
 
     fn read_2d_image<P>(path: P) -> Array2<f32>
-        where P: AsRef<Path>
+    where
+        P: AsRef<Path>,
     {
         let nifti_object = InMemNiftiObject::from_file(path).expect("Nifti file is unreadable.");
         let volume = nifti_object.into_volume();
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -245,3 +245,129 @@ fn write<A, S, D, W>(
 fn to_bytes<T>(data: &[T], size: usize) -> &[u8] {
     unsafe { from_raw_parts::<u8>(data.as_ptr() as *const u8, size) }
 }
+
+#[cfg(test)]
+pub mod tests {
+    extern crate tempfile;
+
+    use super::*;
+
+    use ndarray::{Array, Array2, Ix2, IxDyn, ShapeBuilder};
+    use self::tempfile::tempdir;
+
+    use {InMemNiftiObject, IntoNdArray, header::MAGIC_CODE_NIP1, object::NiftiObject};
+
+    fn get_random_path(ext: &str) -> String {
+        let dir = tempdir().unwrap();
+        let path = if ext == "" {
+            dir.into_path()
+        } else {
+            dir.into_path().join(ext)
+        };
+        path.to_str().unwrap().to_string()
+    }
+
+    pub fn generate_nifti_header(
+        dim: [u16; 8],
+        scl_slope: f32,
+        scl_inter: f32,
+        datatype: i16
+    ) -> NiftiHeader {
+        let bitpix = (mem::size_of::<f32>() * 8) as i16;
+        let magic = *MAGIC_CODE_NIP1;
+        NiftiHeader {
+            dim, datatype, bitpix, magic, scl_slope, scl_inter, ..NiftiHeader::default()
+        }
+    }
+
+    fn read_2d_image(path: &str) -> Array2<f32> {
+        let nifti_object = InMemNiftiObject::from_file(path).expect("Nifti file is unreadable.");
+        let volume = nifti_object.into_volume();
+        let dyn_data = volume.into_ndarray().unwrap();
+        dyn_data.into_dimensionality::<Ix2>().unwrap()
+    }
+
+    fn test_write_read(arr: &Array<f32, IxDyn>, path: &str) {
+        let path = get_random_path(path);
+        let mut dim = [1; 8];
+        dim[0] = arr.ndim() as u16;
+        for (i, s) in arr.shape().iter().enumerate() {
+            dim[i + 1] = *s as u16;
+        }
+        let header = generate_nifti_header(dim, 1.0, 0.0, 16);
+        write_nifti(&path, &arr, Some(&header));
+
+        let read_nifti = read_2d_image(&path);
+        assert!(read_nifti.all_close(&arr, 1e-10) == true);
+    }
+
+    fn f_order_array() -> Array<f32, IxDyn> {
+        let dim = vec![4, 4];
+        let vec = (0..16).map(|x| x as f32).collect();
+        Array::from_shape_vec(IxDyn(&dim).f(), vec).unwrap()
+    }
+
+    fn c_order_array() -> Array<f32, IxDyn> {
+        let dim = vec![4, 4];
+        let vec = (0..16).map(|x| x as f32).collect();
+        Array::from_shape_vec(IxDyn(&dim), vec).unwrap()
+    }
+
+    #[test]
+    fn test_fortran_writing() {
+        // Test .nii
+        let arr = f_order_array();
+        test_write_read(&arr, "test.nii");
+        let mut arr = f_order_array();
+        arr.invert_axis(Axis(1));
+        test_write_read(&arr, "test_non_contiguous.nii");
+
+        // Test .nii.gz
+        let arr = f_order_array();
+        test_write_read(&arr, "test.nii.gz");
+        let mut arr = f_order_array();
+        arr.invert_axis(Axis(1));
+        test_write_read(&arr, "test_non_contiguous.nii.gz");
+    }
+
+    #[test]
+    fn test_c_writing() {
+        // Test .nii
+        let arr = c_order_array();
+        test_write_read(&arr, "test.nii");
+
+        let mut arr = c_order_array();
+        arr.invert_axis(Axis(1));
+        test_write_read(&arr, "test_non_contiguous.nii");
+
+        // Test .nii.gz
+        let arr = c_order_array();
+        test_write_read(&arr, "test.nii.gz");
+
+        let mut arr = c_order_array();
+        arr.invert_axis(Axis(1));
+        test_write_read(&arr, "test_non_contiguous.nii.gz");
+    }
+
+    #[test]
+    fn test_header_slope_inter() {
+        use std::ops::{Add, Mul};
+
+        let arr = f_order_array();
+        let slope = 2.2;
+        let inter = 101.1;
+
+        let path = get_random_path("test_slope_inter.nii");
+        let mut dim = [1; 8];
+        dim[0] = arr.ndim() as u16;
+        for (i, s) in arr.shape().iter().enumerate() {
+            dim[i + 1] = *s as u16;
+        }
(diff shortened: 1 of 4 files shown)


src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+        ..reference
+    };
+
+    let f = File::create(path).expect("Can't create new nifti file");
+    let mut writer = BufWriter::new(f);
+    if is_gz_file(&path) {
+        let mut buffer = Vec::new();
+        write(&mut buffer, header, data);
+
+        let mut e = GzEncoder::new(Vec::new(), Compression::default());
```
Enet4: Why isn't the `GzEncoder` applied directly on the file writer? I would imagine that the gzip encoder already works well for streamed data.

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+
+/// Write a nifti file (.nii or .nii.gz).
+///
+/// If a `reference` is given, it will be used to fill most of the header's fields. The voxels
+/// intensity will be subtracted by `scl_slope` and divided by `scl_inter`. If `reference` is not
+/// given, a default `NiftiHeader` will be built and written.
+///
+/// In all cases, the `dim`, `datatype` and `bitpix` fields will depend only on `data`, not on the
+/// header. In other words, the `datatype` defined in `reference` will be ignored.
+pub fn write_nifti<A, S, D>(
```
Enet4: Errors emerging from this function should be recoverable. In this case, changing the return type to `Result<(), NiftiError>` and raising errors with `?` should be enough.

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+        e.write_all(&buffer).unwrap();
+
+        let buffer = e.finish().unwrap();
+        writer.write_all(&buffer).unwrap();
+    } else {
+        write(&mut writer, header, data);
+    }
+}
+
+fn write<A, S, D, W>(
```
Enet4: As before, this function should have recoverable errors.

The logic for writing the header should be moved to a separate function, so that it can be reused later.

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+            let arr_data = arr_data.into_shape(arr_len).unwrap();
+            let arr_data = arr_data.as_slice().unwrap();
+
+            let buffer = to_bytes(&arr_data, nb_bytes / dim0_size);
+            writer.write_all(&buffer).unwrap();
+        }
+    }
+}
+
+fn to_bytes<T>(data: &[T], size: usize) -> &[u8] {
```
Enet4: `safe-transmute` has a function for this: [`guarded_transmute_to_bytes_pod_many`](https://docs.rs/safe-transmute/0.8.0/safe_transmute/fn.guarded_transmute_to_bytes_pod_many.html)

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+    extern crate tempfile;
+
+    use super::*;
+
+    use ndarray::{Array, Array2, Ix2, IxDyn, ShapeBuilder};
+    use self::tempfile::tempdir;
+
+    use {InMemNiftiObject, IntoNdArray, header::MAGIC_CODE_NIP1, object::NiftiObject};
+
+    fn get_random_path(ext: &str) -> String {
```
Enet4: This could return a `PathBuf`.

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+        datatype: i16
+    ) -> NiftiHeader {
+        let bitpix = (mem::size_of::<f32>() * 8) as i16;
+        let magic = *MAGIC_CODE_NIP1;
+        NiftiHeader {
+            dim, datatype, bitpix, magic, scl_slope, scl_inter, ..NiftiHeader::default()
+        }
+    }
+
+    fn read_2d_image(path: &str) -> Array2<f32> {
```
Enet4: The usual pattern for taking a file path is `P where P: AsRef<Path>`.

```suggestion
    fn read_2d_image<P>(path: P) -> Array2<f32>
    where
        P: AsRef<::std::path::Path>,
    {
```

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+        let mut dim = [1; 8];
+        dim[0] = arr.ndim() as u16;
+        for (i, s) in arr.shape().iter().enumerate() {
+            dim[i + 1] = *s as u16;
+        }
+        let header = generate_nifti_header(dim, 1.0, 0.0, 16);
+        write_nifti(&path, &arr, Some(&header));
+
+        let read_nifti = read_2d_image(&path);
+        assert!(read_nifti.all_close(&arr, 1e-10) == true);
```
Enet4: That `== true` is unnecessary IMO.
```suggestion
        assert!(read_nifti.all_close(&arr, 1e-10));
```

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+        dim[0] = arr.ndim() as u16;
+        for (i, s) in arr.shape().iter().enumerate() {
+            dim[i + 1] = *s as u16;
+        }
+        let header = generate_nifti_header(dim, slope, inter, 16);
+        let transformed_data = arr.mul(slope).add(inter);
+        write_nifti(&path, &transformed_data, Some(&header));
+
+        let read_nifti = read_2d_image(&path);
+        assert!(read_nifti.all_close(&transformed_data, 1e-10) == true);
```
Enet4:
```suggestion
        assert!(read_nifti.all_close(&transformed_data, 1e-10));
```

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+        writer.write_f32::<B>(*f).unwrap();
+    }
+    for f in header.srow_x.iter().chain(&header.srow_y).chain(&header.srow_z) {
+        writer.write_f32::<B>(*f).unwrap();
+    }
+    writer.write_all(&header.intent_name).unwrap();
+    writer.write_all(&header.magic).unwrap();
+
+    // Empty 4 bytes after the header
+    writer.write_u32::<B>(0).unwrap();
```
Enet4: Eventually, we ought to support writing extension data as well. We can keep this gap documented in the mean time.

src/writer.rs
```diff
@@ -0,0 +1,373 @@
...
+use flate2::write::GzEncoder;
+use ndarray::{ArrayBase, Axis, Data, Dimension, RemoveAxis, ScalarOperand};
+use num_traits::FromPrimitive;
+
+use {NiftiHeader, util::is_gz_file};
+
+type B = LittleEndian;
+
+/// Specific actions and header information that depend on the ndarray type.
+pub trait TypeTool<T> {
```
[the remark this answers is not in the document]
nilgoyette: Excellent idea! I am happy to get rid of my trait. Thank you.

src/writer.rs
nilgoyette: Hahaha, wow, I wasn't aware it could do that. Thank you.

src/writer.rs
```diff
@@ -119,168 +52,185 @@ pub fn write_nifti<A, S, D>(
...
-    fn get_random_path(ext: &str) -> String {
+    fn get_random_path(ext: &str) -> PathBuf {
         let dir = tempdir().unwrap();
         let path = if ext == "" {
             dir.into_path()
         } else {
             dir.into_path().join(ext)
         };
-        path.to_str().unwrap().to_string()
+        PathBuf::from(path.to_str().unwrap())
```
Enet4: Isn't `path` already a `PathBuf`? It should be, according to [`into_path()`](https://docs.rs/tempfile/3.0.4/tempfile/struct.TempDir.html#method.into_path).

src/writer.rs
```diff
@@ -119,168 +52,185 @@ pub fn write_nifti<A, S, D>(
...
     use self::tempfile::tempdir;
+    use ndarray::{Array, Array2, Ix2, IxDyn, ShapeBuilder};
 
-    use {InMemNiftiObject, IntoNdArray, header::MAGIC_CODE_NIP1, object::NiftiObject};
+    use {header::MAGIC_CODE_NIP1, object::NiftiObject, InMemNiftiObject, IntoNdArray};
 
-    fn get_random_path(ext: &str) -> String {
+    fn get_random_path(ext: &str) -> PathBuf {
         let dir = tempdir().unwrap();
         let path = if ext == "" {
```
Enet4: With a mut variable, we can improve this a bit:

```rust
let mut path = dir.into_path();
if !ext.is_empty() {
    path.push(ext);
}
```

src/writer.rs
```diff
@@ -0,0 +1,321 @@
...
+            datatype,
+            bitpix: (mem::size_of::<f32>() * 8) as i16,
+            magic: *MAGIC_CODE_NIP1,
+            scl_slope,
+            scl_inter,
+            ..NiftiHeader::default()
+        }
+    }
+
+    fn read_2d_image<P: AsRef<::std::path::Path>>(path: P) -> Array2<f32> {
```
Enet4: My convention is to use `where` clauses for all traits (except `?Sized` and lifetime constraints):
```suggestion
    fn read_2d_image<P>(path: P) -> Array2<f32>
    where
        P: AsRef<::std::path::Path>,
    {
```

nilgoyette: I probably concluded that you put them 'in' when I checked `extension.rs` by "chance"

    pub fn from_stream<S: Read>(mut source: S) -> Result<Self> {

There are some other examples in your code, but I won't touch them because they are not related at all to this PR. I'll change mine.

Enet4: Hmm yeah, that code was written before I left my wicked ways. :blush:

nilgoyette: Hahaha, sorry, I just need to add another example that I just found

    fn from_file_2<P: AsRef<Path>, S>(path: P, mut stream: S) -> Result<InMemNiftiObject>
    where
        S: Read,

Ok, I'll see myself out ^^

src/writer.rs
```diff
@@ -230,7 +232,9 @@ pub mod tests {
         }
     }
 
-    fn read_2d_image<P: AsRef<::std::path::Path>>(path: P) -> Array2<f32> {
+    fn read_2d_image<P>(path: P) -> Array2<f32>
+        where P: AsRef<Path>
```
Enet4: Is that how `rustfmt` put it?

src/writer.rs
Enet4: Hah, I figured there'd be more. This one [here](https://github.com/Enet4/nifti-rs/blob/master/src/extension.rs#L136) is worse though, as it clearly shows how adding the type constraints at the beginning makes a long-cat long function prototype. If we ever consider a full restyle, this one ought to go first.

```rust
    pub fn from_stream<B: ByteOrder, S: Read>(
        extender: Extender,
        mut source: S,
        len: usize,
) -> Result<Self> {
```

src/writer.rs
nilgoyette: My bad, sorry. I'm not used to run `rustfmt`! Will do.

PR #23, opened 2018-11-07
Use Compression::fast

nilgoyette:
We wondered for a long time why our writer was much slower than NiBabel. Turns out that `Compression::default()` is not

1. fast at all
2) what NiBabel [does](http://nipy.org/nibabel/reference/nibabel.openers.html). It's misleading. It looks like they use 9 (slowest) but their default is actually 1 (fastest).

With that change, we are as fast and even faster than NiBabel.


nilgoyette:
I forgot. Of course, we could make this configurable.

PR #26, opened 2018-11-09
RGB images writer

nilgoyette:
There's some code copy, but I don't know how to avoid it. Most types can have a `slope` so the standard functions respect:

     where
        ...
        A: Div<Output = A>,
        A: FromPrimitive,
        A: ScalarOperand,
        A: Sub<Output = A>
        ...

Of course, `[u8; 3]` can't respect those traits, so I added `write_rgb_nifti`, which is almost identical to `write_nifti`, and I try to re-use utility functions as much as possible. 

Moreover, I added 2 tests where I write RGB images, but I can't read them... So, those tests only check that the code compiles I guess. We will need a RGB reader someday to fix this!


src/writer.rs
```diff
@@ -149,51 +212,61 @@ where
 /// Write the data in 'f' order.
 ///
 /// Like NiBabel, we iterate by "slice" to improve speed and use less memory.
-fn write_data<A, S, D, W>(writer: &mut W, header: NiftiHeader, data: &ArrayBase<S, D>) -> Result<()>
+fn write_data<T, D, W>(writer: &mut W, header: NiftiHeader, data: ArrayView<T, D>) -> Result<()>
```
Enet4: `header` can be passed by reference, since it's not meant to be consumed here.

src/writer.rs
```diff
@@ -149,51 +212,61 @@ where
...
     let slope = if header.scl_slope == 0.0 {
         1.0
     } else {
         header.scl_slope
     };
     if slope != 1.0 || header.scl_inter != 0.0 {
-        let slope = A::from_f32(slope).unwrap();
-        let inter = A::from_f32(header.scl_inter).unwrap();
+        let slope = T::from_f32(slope).unwrap();
+        let inter = T::from_f32(header.scl_inter).unwrap();
```
Enet4: I think this part of the function should be reiterated (either now or in a separate PR), since it does not exactly resemble the current logic employed when reading transformed data. The [`data::element`](https://github.com/Enet4/nifti-rs/blob/master/src/volume/element.rs#L20) module provides multiple linear transformation strategies for each valid element data type to prevent precision loss. In this particular code, an `scl_slope` of `0.5` (haven't even seen real data with this, but it's a valid one nevertheless) would turn all voxel values to 0 if we pass an ndarray of integers. We might need a bit more code than just reusing this module though, since the transformation strategies currently live in immaterial types.

src/writer.rs
```diff
@@ -149,51 +212,61 @@ where
...
+
+fn write_slices<T, D, W>(writer: &mut W, data: ArrayView<T, D>) -> Result<()>
+where
+    T: Clone + PodTransmutable,
+    D: Dimension + RemoveAxis,
+    W: WriteBytesExt,
+{
+    for arr_data in data.axis_iter(Axis(0)) {
+        // We need to own the data because of the into_shape() for `C` ordering.
+        write_slice(writer, arr_data.to_owned())?;
```
Enet4: Maybe we don't have to copy data here. Try passing an `ArrayView` instead and making `write_slice` accept any `ArrayBase<S, D>`.

src/writer.rs
```diff
@@ -77,6 +62,84 @@ where
...
+pub fn write_rgb_nifti<P, D>(
+    path: P,
+    data: &Array<[u8; 3], D>,
+    reference: Option<&NiftiHeader>,
+) -> Result<()>
+where
+    P: AsRef<Path>,
+    D: Dimension + RemoveAxis,
+{
+    // Copy most of the reference header, but the `bitpix`, `scl_slope` and `scl_inter` fields must
```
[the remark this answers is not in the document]
nilgoyette: Good question. I choose those 3 because they were necessary. If I remove the UNUSED and all fields related to the affine, we have `intent` stuff, `dim_info`, slice_start/end/code, cal_min/max, `slice_duration`, `toffset` and `descrip`.

- `intent_code` and friends should be set to `0` I think. It would be `RGB: 2003` if we used the "rgb triplet in the 5th dimension", but we don't. I thought this was a hack but it's actually recognized by the nifti standard.
- `cal_min` and `cal_max` probably don't make much sense for a RGB image.
- `toffset` should always be set to 0.0 for 3D images, I guess. But it looks like an unimportant field.
- `descrip`, we just can't know.

The user can still give a reference header if he wants to modify some fields. So, do you want me to change something?

src/writer.rs
nilgoyette: Hum, ok? I did it, no problem, but can you please explain why this should be done? I had a "pragmatic" view: "I don't need it anymore, lets pass it by value", but this may not be idiomatic to Rust. I had some problems in other projects where some callers want to pass by value and others by reference, so I didn't know what to do.

By "it's not meant to be consumed", there's clearly an intent. If the function is calling `into_*` or an object's life should clearly end there, it should be pass by value?

src/writer.rs
nilgoyette: Ouch. True. I'll check if I can copy your design.

src/writer.rs
nilgoyette: Without the copy the image will be flipped. The comment could probably be made clearer :) Both writing functions start with 

     // Need the transpose for fortran ordering used in nifti file format.
     let data = data.t();

but we know that `ndarray` doesn't actually flip the data, it simply changes the strides. The copy ensures that the data in memory is ordered so it can be fed to `guarded_transmute_to_bytes_pod_many`.

What I don't understand (a part of the code was made by a colleague who doesn't recall everything) is why we **always** need to flip. It makes sense only if `ndarray` doesn't actually care about 'c' or 'f' ordering and allocates your data in 'c' anyway. That's probably the case because our tests work!

src/writer.rs
Enet4: Long answer: This isn't a hard rule, since it's a private function and will hardly influence how the program works. My perception of this is that a reference to the header is reasonable from both ends of the spectrum: on the implementation side, the function does need to own any of its parts, nor does it seem that it could take advantage of owning the object. On the interface side, the header provides a piece of processed information to produce an output through a writer channel, and so will have to replicate it into the byte stream regardless. As far as I know in terms of API conventions, the fact that it's a terminal operation doesn't mean that it should consume its operands. As an example, encoding functions in serde usually receive the serializable object by reference.

The short answer is that Clippy has a default lint for this. 😅 The lint itself, `needless_pass_by_value`, has a known history of lame false positives, but here a trigger is reasonable (for the reasons above).

src/writer.rs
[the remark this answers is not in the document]
Enet4: Those don't pose an issue for the time being, I was mostly concerned about the three fields I mentioned above. Unless I overlooked it, these ones can still be overridden by the caller, and I can only picture this as room for disasters with no legitimate use cases.

`magic` should be set to the proper code if it's not already correct (`ni1\0` for headers, `n+1\0` for full volumes). `sizeof_hdr` should be always 348. And for now (since we don't support writing extensions), `vox_offset` should be always 352.

src/writer.rs
Enet4: All right, this is interesting, to say the least. It seems reasonable for `to_owned` to just ignore the ordering of the original volume and arrange it in the default row-major ordering. In turn, the `into_shape` which flattens the array (line 264) will just discard the previously establish F order again.

Well, here's what we should do: we want to preserve memory usage while benefiting from fast writing of contiguous byte sections, so let's allocate a single array in C order and reuse it for every slice. Here's a patch with the trick:

```patch
diff --git a/src/writer.rs b/src/writer.rs
index 3a6cf46..f7a2b59 100644
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -241,22 +241,31 @@ where
     Ok(())
 }
 
-fn write_slices<T, D, W>(writer: &mut W, data: ArrayView<T, D>) -> Result<()>
+fn write_slices<A, S, D, W>(writer: &mut W, data: ArrayBase<S, D>) -> Result<()>
 where
-    T: Clone + PodTransmutable,
+    S: Data<Elem = A>,
+    A: Clone + PodTransmutable,
     D: Dimension + RemoveAxis,
     W: WriteBytesExt,
 {
-    for arr_data in data.axis_iter(Axis(0)) {
-        // We need to own the data because of the into_shape() for `C` ordering.
-        write_slice(writer, arr_data.to_owned())?;
+    let mut iter = data.axis_iter(Axis(0));
+    if let Some(arr_data) = iter.next() {
+        // keep slice voxels in a separate array to ensure `C` ordering
+        // even after `into_shape`.
+        let mut slice = arr_data.to_owned();
+        write_slice(writer, slice.view())?;
+        for arr_data in iter {
+            slice.assign(&arr_data);
+            write_slice(writer, slice.view())?;
+        }
     }
     Ok(())
 }
 
-fn write_slice<T, D, W>(writer: &mut W, data: Array<T, D>) -> Result<()>
+fn write_slice<A, S, D, W>(writer: &mut W, data: ArrayBase<S, D>) -> Result<()>
 where
-    T: Clone + PodTransmutable,
+    S: Data<Elem = A>,
+    A: Clone + PodTransmutable,
     D: Dimension,
     W: WriteBytesExt,
 {
```

src/writer.rs
```diff
@@ -324,4 +397,30 @@ pub mod tests {
...
+    #[test]
+    fn test_write_3d_rgb() {
+        let mut data = Array::from_elem((10, 10, 8), [255u8, 255u8, 255u8]);
+        data[(5, 5, 2)] = [55, 55, 0];
+        data[(5, 5, 3)] = [55, 0, 55];
+        data[(5, 5, 4)] = [0, 55, 55];
+
+        let path = get_temporary_path("rgb.nii.gz");
+        write_rgb_nifti(path, &data, None).unwrap();
+        // TODO Don't just test if the writing compiles and runs, we also need to read it back
```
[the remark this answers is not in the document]
nilgoyette: I'm curiously unable to obtain an gzip identical to NiBabel. My .nii image is identical to NiBabel and vtk, but the gzip aren't. The problem is that I would prefer uploading a gz image on git. Is it a problem for you if I use our own output as the groundtruth? I mean, it looks kinda bad when you know it!

If we ever change the compression factor or compression lib, we will need to remplace the image though. Maybe it's better to have a 433 bytes 3x3x3 image?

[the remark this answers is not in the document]
Enet4: Hmm yes, I guess it's better to reduce the volume's size and hold the ground truth uncompressed. Hopefully the output should be reproducible then.

nilgoyette:
TODO, in another PR, if you agree

- Support hdr writing
- Inverse Linear Transform

Issue #27, opened 2018-11-21
minimal_by_hdr_and_img_gz fails on Windows


nilgoyette:
This is not a new problem, I just waited a long time to report it.

When I run `cargo test --release --features ndarray_volumes`, `minimal_by_hdr_and_img_gz` fails on Windows and WSL. Moreover, it stops the other tests from running. To see the tests from `volume.rs` and `writer.rs`, I must remove the `minimal_by_hdr_and_img_gz` test.

I didn't investigate yet but I'm probably the only one with Windows so I'll do that "soon". What we could do to confirm the problem is to tell travis to test all OS. They added Windows support in the last months.

    os:
      - windows
      - linux
      - osx
    language: rust
    ...

Enet4:
Is this a different issue from #4? But yes, whatever the problem is here, we could add osx and windows to Travis.

nilgoyette:
Hahaha, wow, I totally forgot about this issue. I'll confirm that it has the same cause and, if so, I'll fix the tests. We can't really ask other contributors or travis to use the [hack](https://fishingcatblog.wordpress.com/2016/06/01/symbolic-links-in-windows-git-and-atom/) you suggested.

Enet4:
Closing as fixed by #29.

PR #29, opened 2018-11-23
Remove symlinks and test all OS

### LINKED ISSUE #4: Windows support
I've been using your library on Windows and Debian an everything works well except one test. `minimal_by_hdr_and_img_gz()` uses `resources/minimal2.hdr` which is a link to `minimal.hdr`. Of course, Windows doesn't support link so it's trying to read the link as if it was the actual header.

I hesitate to change it myself because I don't know why you choosed a link in the first place. It seems to me that this test would work as much using `resources/minimal.hdr` but you already have this test so you must be testing something else... Anyway, I can fix it if you can't test on windows.

nilgoyette:
Should fix #4 and #27.


nilgoyette:
Haha, wow, I broke linux and macOS but Windows is working now :D I think `minimal2.hdr` and `minimal2.img.gz` are still considered symlinks for some reasons. I'll try removing and adding the files instead of replacing them.

PR #31, opened 2018-11-27
WIP .hdr writer

nilgoyette:
WIP because `write_nifti` is so ugly now that I didn't found the courage to make `write_rgb_nifti` work too. I would be really interested in your review. Like, how I can make this less ugly. Or even beautiful!

I added a lot of `#[cfg(feature = "ndarray_volumes")]` for `is_hdr_file` because it's currently only used with `ndarray_volumes`, but this will be removed "soon" when we support writing nifti-rs types.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -3,7 +3,7 @@
 use std::fs::File;
 use std::io::BufWriter;
 use std::ops::{Div, Sub};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use byteorder::{LittleEndian, WriteBytesExt};
 use flate2::write::GzEncoder;
@@ -13,8 +13,10 @@ use num_traits::FromPrimitive;
 use safe_transmute::{guarded_transmute_to_bytes_pod_many, PodTransmutable};
 
 use {
-    header::MAGIC_CODE_NIP1, util::is_gz_file, volume::element::DataElement, NiftiHeader,
-    NiftiType, Result,
+    header::{MAGIC_CODE_NIP1, MAGIC_CODE_NI1},
+    util::{is_gz_file, is_hdr_file},
+    volume::element::DataElement,
+    NiftiHeader, NiftiType, Result,
 };
 
 // TODO make this configurable. The Nifti standard does not specify a specific field for endianness,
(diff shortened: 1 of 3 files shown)


src/writer.rs
```diff
@@ -140,7 +168,24 @@ where
         magic: *MAGIC_CODE_NIP1,
         // All other fields are copied from reference header
         ..reference
-    }
+    };
+
+    let mut path_buf = PathBuf::from(path.as_ref());
+    let (header_path, data_path) = if is_hdr_file(&path) {
```
Enet4: Isn't `header_path = path` regardless of which branch is taken?

src/writer.rs
```diff
@@ -46,22 +48,42 @@ where
...
 
-    let f = File::create(&path)?;
-    let mut writer = BufWriter::new(f);
-    if is_gz_file(&path) {
-        let mut e = GzEncoder::new(writer, Compression::fast());
-        write_header(&mut e, &header)?;
-        write_data(&mut e, &header, data)?;
-        let _ = e.finish()?; // Must use result
+    let header_file = File::create(header_path)?;
+    let mut header_writer = BufWriter::new(header_file);
```
Enet4: Since `GzEncoder` already uses a buffer for the gz encoding process, wouldn't it be best to leave `BufWriter` out if `is_gz`?

src/writer.rs
```diff
@@ -46,22 +48,42 @@ where
...
+            let _ = e.finish()?;
+        } else {
+            write_header(&mut header_writer, &header)?;
+            write_data(&mut header_writer, &header, data)?;
+        }
     } else {
-        write_header(&mut writer, &header)?;
-        write_data(&mut writer, &header, data)?;
+        let data_file = File::create(&data_path)?;
+        let mut data_writer = BufWriter::new(data_file);
```
Enet4: Same concern: gz encoding is inherently buffered.

tests/writer.rs
```diff
@@ -87,7 +87,7 @@ mod tests {
     }
 
     #[test]
```
Enet4: Would it be worth making another test for a non-contiguous array, that is in neither C nor F order?

src/writer.rs
nilgoyette: TIL. Yes, sure.

tests/writer.rs
nilgoyette: Excellent idea. I'll add that to my list and contribute it in another PR.

Issue #34, opened 2018-12-10
Data writing is always done in native byte order regardless of header


Enet4:
If the target's endianness is Big Endian, the voxel data will be stored in the system's native order, while the header is stored in little endian. This will make the header/volume pair inconsistent, leading to garbage voxel data.

This can be reproduced by building and testing for the `mips64-unknown-linux-gnuabi64` target.

```
cross test --features ndarray_volumes --target mips64-unknown-linux-gnuabi64
```

```
running 5 tests
test tests::test_c_writing ... FAILED
test tests::test_fortran_writing ... FAILED
test tests::test_header_slope_inter ... FAILED
test tests::test_write_3d_rgb ... ok
test tests::test_write_4d_rgb ... ok

failures:

---- tests::test_c_writing stdout ----
thread 'tests::test_c_writing' panicked at 'assertion failed: read_nifti.all_close(&arr, 1e-10)', tests/writer.rs:74:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- tests::test_fortran_writing stdout ----
thread 'tests::test_fortran_writing' panicked at 'assertion failed: read_nifti.all_close(&arr, 1e-10)', tests/writer.rs:74:9

---- tests::test_header_slope_inter stdout ----
thread 'tests::test_header_slope_inter' panicked at 'assertion failed: read_nifti.all_close(&transformed_data, 1e-10)', tests/writer.rs:142:9


failures:
    tests::test_c_writing
    tests::test_fortran_writing
    tests::test_header_slope_inter

test result: FAILED. 2 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out
```

The best way to fix this is to write the header in exactly the same endianness as the `header.endianness`, and then write the volume data in the same byte order. We can still use `write_all` for each slice if we first reverse the bytes of each voxel value in case of a non-native order.

I will be adjusting our Travis CI job matrix to also cover big endian systems, as we really have a fair amount of code that depends on the target system's byte order (while taking away some other less important jobs).

nilgoyette:
I consider `header.endianness` as a general indication on the endianness of the file

- not reliable because it could disagree with the data. I don't have such data but I'm pretty sure it exists. I read somewhere that the standard "encoded in `dim[0]`. if `dim[0]` is outside range 1..7, then swap".
- not as an actual order from the programmer who sent the header.

[...]

PR #37, opened 2018-12-19
Use byteordered

Enet4:
This PR changes the data encoding and decoding routines to use [`byteordered`](https://docs.rs/byteordered/0.3.1/byteordered/). It is an alternative API with a byte order wrapper, so that reads and writes do not have to explicitly state the expected order in each call, and it also works for an endianness only known at run-time.

This makes the code base a bit cleaner, especially in util.rs since we no longer have to define `Endianness` here, and should also make #34 significantly easier to address. The only breaking change is that now `Endianness` is provided by `byteordered`, which has a few changes in associated functions and variant names (conventional naming: `Little` instead of `LE` and `Big` instead of `BE`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ branch = "master"
 repository = "Enet4/nifti-rs"
 
 [dependencies]
-byteorder = "1.2.1"
+byteordered = "0.3.1"
 derive_builder = "0.7.0"
 flate2 = "1.0.1"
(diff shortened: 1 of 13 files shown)


nilgoyette:
Hey, thank you for this nice lib. I'll check if it makes sense to use it in my trk lib.

I didn't read everything carefully because it's all mixed in formatting changes, but I do agree that's it's cleaner.

PR #42, opened 2019-01-30
Affine management

### LINKED ISSUE #22: Affine transformation
We would like to contribute "affine" code to nifti-rs. A part of the code is in our private compagnie codebase and the rest is [here](https://github.com/imeka/trk-io/blob/master/src/affine.rs).

I think it belongs in nifti-rs because:

- it doesn't belong in trk-io, nor in our private codebase :)
- it totally depends on `NiftiHeader` struct.
- some people might need it and this is where they will search first.

In fact, I consider nifti-rs to be the Rust version of NiBabel, so of course it should be in nifti-rs. What's your opinion on this?

Disclaimer: Most of the code has been **ported** from NiBabel. I don't know much about licensing and I don't like to care about it.

nilgoyette:
Fix #22. Adds 2 public methods to `NiftiHeader` and a lot of other methods/functions in `header.rs` and `affine.rs`. This is mostly a port from NiBabel because:

- I don't actually know how quaternion works
- NiBabel has been tested extensively, by us and the whole community.

We (at Imeka) have been testing this port for more than 6 months now and we're quite happy with it. The code was in our open sourced project called [trk-io](https://github.com/imeka/trk-io/blob/master/src/affine_nifti.rs). I simply cleaned it, moved some functions to methods, etc.

I do have some questions though:

- I'm still unsure what should be public. `get_affine` and `set_affine` is the bare minimum and this is what I choosed. It might be a good idea to set the 'sform' and 'qform' methods public too.
- You know from my previous contributions that I don't use `Error` much. I prefer the `panic` alternative, but I know you will tell me to use `Error` :)
- I did everything in f32 but it should either be templated (to avoid forcing the users to convert their matrices) or in f64 (the math, at least). I'll probably do that in my next version.
- I hate formatting when I use matrices! Continued below.

I know I can use `#[rustfmt::skip]` on construction, but I don't like to see those things in my code. And you can't use it when you return a variable :| I tried using `from_rows` to avoid long lines but some lines in some matrices are still too long. I simply called rustfmt on `header.rs` and `affine.rs`, then I reverted the changes on the matrices because it's unreadable. You might disagree on this subject, I'm interested to know your opition.

    Affine4::from_rows(
        Vector4::new(m[0] as f32, m[3] as f32, m[6] as f32, self.quatern_x),
        Vector4::new(m[1] as f32, m[4] as f32, m[7] as f32, self.quatern_y),
        Vector4::new(m[2] as f32, m[5] as f32, m[8] as f32, self.quatern_z),
        Vector4::new(0.0, 0.0, 0.0, 1.0),
    )

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -139,3 +139,63 @@ pub(crate) fn quaternion_to_affine(q: Quaternion<f32>) -> Affine3 {
         xz - wy, yz + wx, 1.0 - (xx + yy),
     )
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_shape_zoom_affine() {
+        let affine = shape_zoom_affine(&[3, 5, 7], &[3.0, 2.0, 1.0]);
+        assert_eq!(affine, Affine4::new(
+            -3.0, 0.0, 0.0, 3.0,
+            0.0, 2.0, 0.0, -4.0,
+            0.0, 0.0, 1.0, -3.0,
+            0.0, 0.0, 0.0, 1.0,
+        ));
+
+        let affine = shape_zoom_affine(&[256, 256, 54], &[0.9375, 0.9375, 3.0]);
+        assert_eq!(affine, Affine4::new(
+            -0.9375, 0.0, 0.0, 119.53125,
+            0.0, 0.9375, 0.0, -119.53125,
+            0.0, 0.0, 3.0, -79.5,
+            0.0, 0.0, 0.0, 1.0,
+        ));
+    }
+
+    #[test]
+    fn test_fill_positive() {
+        let q = fill_positive(Vector3::new(0.0, 0.0, 0.0));
+        assert_eq!(q, Quaternion::new(1.0, 0.0, 0.0, 0.0));
+
+        let q = fill_positive(Vector3::new(1.0, 0.0, 0.0));
+        assert_eq!(q, Quaternion::new(0.0, 1.0, 0.0, 0.0));
+        assert_eq!(q.dot(&q), 1.0);
+    }
+
+    #[test]
+    fn test_affine_to_quaternion() {
+        let affine = Matrix3::<f32>::identity();
+        assert_eq!(affine_to_quaternion(&affine), RowVector4::new(1.0, 0.0, 0.0, 0.0));
+
+        let affine = Matrix3::from_diagonal(&Vector3::new(1.0, -1.0, -1.0));
+        assert_eq!(affine_to_quaternion(&affine), RowVector4::new(0.0, 1.0, 0.0, 0.0));
+
+        let affine = Matrix3::new(1.1, 0.1, 0.1, 0.2, 1.1, 0.5, 0.0, 0.0, 1.0);
+        assert_eq!(
+            affine_to_quaternion(&affine),
+            RowVector4::new(0.99299955, -0.1147423, 0.017765911, 0.021675035)
+        );
+    }
+
+    #[test]
+    fn test_quaternion_to_affine() {
+        // Identity quaternion
+        let affine = quaternion_to_affine(Quaternion::new(1.0, 0.0, 0.0, 0.0));
+        assert_eq!(affine, Matrix3::identity());
+
+        // 180 degree rotation around axis 0
+        let affine = quaternion_to_affine(Quaternion::new(0.0, 1.0, 0.0, 0.0));
+        assert_eq!(affine, Matrix3::new(1.0, 0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0, -1.0));
+    }
+}
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -1,12 +1,16 @@
+//! This module defines some affine-related utilities.
+
 use nalgebra::{Matrix3, Matrix4, Quaternion, RowVector4, Scalar, SymmetricEigen, Vector3, U1, U3};
 
+/// 3x3 affine transformation matrix.
 pub type Affine3 = Matrix3<f32>;
+/// 4x4 affine transformation matrix.
 pub type Affine4 = Matrix4<f32>;
 
 const QUARTERNION_THRESHOLD: f32 = -::std::f32::EPSILON * 3.0;
 
 /// Separate a 4x4 affine into its 3x3 affine and translation components.
-pub fn get_affine_and_translation<T: Scalar>(affine: &Matrix4<T>) -> (Matrix3<T>, Vector3<T>) {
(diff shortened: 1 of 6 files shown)


src/affine.rs
```diff
@@ -0,0 +1,212 @@
+use nalgebra::{Matrix3, Matrix4, Quaternion, RowVector4, Scalar, SymmetricEigen, Vector3, U1, U3};
+
+pub type Affine3 = Matrix3<f32>;
+pub type Affine4 = Matrix4<f32>;
+
+const QUARTERNION_THRESHOLD: f32 = -::std::f32::EPSILON * 3.0;
+
+/// Separate a 4x4 affine into its 3x3 affine and translation components.
+pub fn get_affine_and_translation<T: Scalar>(affine: &Matrix4<T>) -> (Matrix3<T>, Vector3<T>) {
```
Enet4: Non-idiomatic `get_` prefix. This operation seems to _convert_ something more than it _gets_ something, and even as a getter, it should not start with `get_` as per the [C-GETTER](https://rust-lang-nursery.github.io/api-guidelines/naming.html#c-getter) guidelines. ¹

src/header.rs
```diff
@@ -324,6 +328,160 @@ impl NiftiHeader {
...
+impl NiftiHeader {
+    /// Retrieve best of available transformations.
+    ///
+    /// Return the 'sform' transformation if `sform_code` has a valid value, 'qform' transformation
+    /// if `qform_code` has a valid value, otherwise return a "base" transformation, constructed
+    /// from the declared shape and zooms.
+    ///
+    /// If `sform_code` and `qform_code` both have valid values, the 'sform' affine transformation
+    /// is prioritized.
+    pub fn get_affine(&self) -> Affine4 {
```
Enet4: Also non-idiomatic `get_` prefix (¹)

src/header.rs
```diff
@@ -324,6 +328,160 @@ impl NiftiHeader {
...
+            self.get_sform_affine()
+        } else if self.qform_code != 0 {
+            self.get_qform_affine()
+        } else {
+            self.get_base_affine()
+        }
+    }
+
+    /// Retrieve affine transformation from 'sform' fields.
+    fn get_sform_affine(&self) -> Affine4 {
```
Enet4: ¹

src/header.rs
```diff
@@ -324,6 +328,160 @@ impl NiftiHeader {
...
+        Affine4::new(
+            self.srow_x[0], self.srow_x[1], self.srow_x[2], self.srow_x[3],
+            self.srow_y[0], self.srow_y[1], self.srow_y[2], self.srow_y[3],
+            self.srow_z[0], self.srow_z[1], self.srow_z[2], self.srow_z[3],
+            0.0, 0.0, 0.0, 1.0,
+        )
+    }
+
+    /// Retrieve affine transformation from qform-related fields.
+    fn get_qform_affine(&self) -> Affine4 {
```
Enet4: ¹

src/header.rs
```diff
@@ -324,6 +328,160 @@ impl NiftiHeader {
...
+            m[1] as f32, m[4] as f32, m[7] as f32, self.quatern_y,
+            m[2] as f32, m[5] as f32, m[8] as f32, self.quatern_z,
+            0.0, 0.0, 0.0, 1.0,
+        )
+    }
+
+    /// Retrieve affine transformation implied by shape and zooms.
+    ///
+    /// Note that we get the translations from the center of the image.
+    fn get_base_affine(&self) -> Affine4 {
```
Enet4: ¹

src/header.rs
```diff
@@ -324,6 +328,160 @@ impl NiftiHeader {
...
+        let d = self.dim[0] as usize;
+        shape_zoom_affine(&self.dim[1..d + 1], &self.pixdim[1..d + 1])
+    }
+
+    /// Compute quaternion from b, c, d of quaternion.
+    ///
+    /// Fills a value by assuming this is a unit quaternion.
+    fn get_qform_quaternion(&self) -> Quaternion<f32> {
```
Enet4: ¹

src/header.rs
```diff
@@ -324,6 +328,160 @@ impl NiftiHeader {
...
+        fill_positive(Vector3::new(self.quatern_b, self.quatern_c, self.quatern_d))
+    }
+
+    /// Set affine transformation.
+    ///
+    /// Will set both affine transformations to avoid interoperability problems:
+    ///
+    /// * 'sform' from unmodified `affine`, with `sform_code` set to `AlignedAnat`.
+    /// * 'qform' from a quaternion built from `affine`. However, the 'qform' won't be used by most
+    /// nifti readers because the `qform_code` will be set to 'Unknown'.
```
Enet4: If by `Unknown` you mean the `XForm` variant, I guess we can do this:

```suggestion
    /// nifti readers because the `qform_code` will be set to `Unknown`.
```

tests/affine.rs
```diff
@@ -0,0 +1,115 @@
...
+        header.srow_y = [0.1, 2.4995277, 0.0485984, -97.420204];
+        header.srow_z = [0.4, -0.0485, 2.4991884, -89.12282];
+
+        let real_affine = Affine4::new(
+            2.4, -0.0008,    -0.0411765, -114.766396,
+            0.1,  2.4995277,  0.0485984,  -97.420204,
+            0.4, -0.0485,     2.4991884,  -89.12282,
+            0.0,  0.0,        0.0,          1.0
+        );
+        assert_eq!(header.affine(), real_affine);
```
Enet4: When performing equality checks between values with floating point numbers, it would be best to perform an approximate equality check (as in, assuming a small value margin, like in most other tests we have right now). Is there something we can use in `nalgebra` for this?

src/lib.rs
```diff
@@ -59,7 +60,7 @@ extern crate flate2;
 extern crate num_traits;
 extern crate safe_transmute;
 
-#[cfg(feature = "nalgebra_affine")] mod affine;
+#[cfg(feature = "nalgebra_affine")] pub mod affine;
```
Enet4: Just raising the question: does it make sense for this module to become part of the crate's public API?

tests/affine.rs
nilgoyette: Yes, they use `approx` in the `partial_cmp` free-[function](https://www.nalgebra.org/rustdoc/src/nalgebra/lib.rs.html#326-328). Will do.

Enet4:
OK, think we're ready to merge this, with the expectation that the module will be useful. If at any time you find yourself using this module for something unrelated to the NIfTI file format, we may then consider moving that logic to a dedicated crate and depend on it here.

PR #43, opened 2019-02-12
Validate dimensionality

Enet4:
I found a problem while fuzzing the crate. If `dim[0]` contains an inappropriate value, attempting to read a volume would panic. This PR makes some changes to accommodate the fix.

- new error variant for inconsistent dim field values (must be positive, and `dim[0]` must not be higher than 7).
- raise this error if `dim[0]` is incorrect.
- as a bonus, I added public methods to `NiftiHeader`, which validate the values in the process.


src/header.rs
```diff
@@ -231,6 +231,31 @@ impl NiftiHeader {
...
+    }
+
+    /// Retrieve and validate the number of dimensions of the volume. This is
+    /// `dim[0]` after the necessary byte order conversions.
+    /// 
+    /// # Error
+    /// 
+    /// `NiftiError::` if `dim[0]` does not represent a valid dimensionality.
+    pub fn dimensionality(&self) -> Result <usize> {
+        if self.dim[0] > 7 {
```
nilgoyette: Is 0 a valid dimensionality?

src/lib.rs
```diff
@@ -47,6 +47,7 @@
 //!
 #![deny(missing_debug_implementations)]
 #![warn(missing_docs, unused_extern_crates, trivial_casts, unused_results)]
+#![recursion_limit = "128"]
```
nilgoyette: Can you please tell me why we need a recursion limit? I don't remember any use of recursion in this lib.

Enet4:
Thanks for reviewing, @nilgoyette. I can double-check those patterns and replace occurrences where reasonable the next few days.

More comments inline.

src/header.rs
Enet4: Great question, I don't know. Might be worth looking into this.

src/lib.rs
Enet4: This is the compiler's recursive operation limit. Recursion is used internally by the `quick_error!` macro. Once a significant number of variants are declared, which is the case here, we need to push this value a bit higher.

nilgoyette:
LGTM.

PR #47, opened 2019-02-28
Add a half_slope test

nilgoyette:
I forgot to add a test in my last review. My test fails (Division by zero) on `6d7c909e12ad` and it passes on master.


nilgoyette:
Thank you. No, there's no gain for the user. It's only useful for us.

PR #48, opened 2019-04-18
Update alga + nalgebra version.

czotti:
We found that `nalgebra` and `alga` can be updated.

- nalgebra 0.18
- alga 0.9
- one test is slightly different (less than 10e-13)
- change deprecated function `into_subview` into `index_axis_move`

 I have made test with these commands:
```
cargo build --release --features ndarray_volumes
cargo test --release --features ndarray_volumes
cargo build --release --features nalgebra_affine
cargo test --release --features nalgebra_affine
cargo build --release --features ndarray_volumes,nalgebra_affine
cargo test --release --features ndarray_volumes,nalgebra_affine
```

everything looks fine.


nilgoyette:
LGTM. Disclaimer: @czotti is my colleague.

src/affine.rs
```diff
@@ -198,10 +198,10 @@ mod tests {
         assert_eq!(
             affine_to_quaternion(&affine),
             RowVector4::new(
```
Enet4: Well, this one could really use an approximate equality fix, so as to avoid having test failures due to floating point arithmetic precision issues. By tapping into `RowVector4`, we can obtain a slice, and from there employ one of the macros in the `approx` crate.

Here's a patch that you can apply to address this:

```diff
diff --git a/src/affine.rs b/src/affine.rs
index e158519..cebe1da 100644
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -195,14 +195,17 @@ mod tests {
         );
 
         let affine = Matrix3::new(1.1, 0.1, 0.1, 0.2, 1.1, 0.5, 0.0, 0.0, 1.0);
-        assert_eq!(
-            affine_to_quaternion(&affine),
-            RowVector4::new(
+
+        let quaternion = affine_to_quaternion(&affine);
+        assert_abs_diff_eq!(
+            quaternion.as_slice(),
+            &[
                 0.9929998817020889,
                 -0.11474227051531193,
                 0.017766153114299018,
-                0.02167510323267152
-            )
+                0.02167510323267152,
+            ][..],
+            epsilon = 1e-11,
         );
     }
 
diff --git a/src/lib.rs b/src/lib.rs
index 5724dc7..05c612f 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,6 +55,7 @@
 #[cfg(feature = "nalgebra_affine")] extern crate alga;
 #[cfg(feature = "nalgebra_affine")] extern crate nalgebra;
 #[cfg(feature = "ndarray_volumes")] extern crate ndarray;
+#[cfg(all(test, feature = "nalgebra_affine"))] #[macro_use] extern crate approx;
 
 extern crate byteordered;
 extern crate flate2;
```

src/volume/ndarray.rs
```diff
@@ -103,6 +103,6 @@ where
     {
         // TODO optimize this implementation (we don't need the whole volume)
         let volume = self.volume.into_ndarray()?;
-        Ok(volume.into_subview(Axis(self.axis as Ix), self.index as usize))
+        Ok(volume.index_axis_move(Axis(self.axis as Ix), self.index as usize))
```
Enet4: Is this change necessary for the updated crates to work? Since this method was only introduced in recent versions of `ndarray`, I was hoping to choose the right moment to make this transition.

czotti: No everything was fine, just a warning during the compilation.
```
warning: use of deprecated item 'ndarray::impl_methods::<impl ndarray::ArrayBase<S, D>>::into_subview': renamed to `index_axis_move`
```

I can revert it if you prefer.

src/affine.rs
czotti: That pretty nice, thanks.

src/volume/ndarray.rs
Enet4: Yes please, if you don't mind. I am aware of this quirk, but changing the method name here requires a different range of compatible `ndarray` versions, and I'd like to record this change separately.

Enet4:
Thanks!

PR #51, opened 2019-06-26
Update safe-transmute to 0.11.0 rc

Enet4:
`safe-transmute` has a history of being a problematic library. This new version of the crate addresses many memory safety issues, making it magnitudes safer to use. Although it is an RC, a migration is still recommended from my point of view.


nilgoyette:
Haha, `transmute_to_bytes` is such a better name!

Issue #44, opened 2019-02-21
Huge gz file


nilgoyette:
We received an big image from the Human Connectome Project, nothing huge, but we needed to resample it to 1x1x1 and now it's 2.3Gb in `.nii.gz` and 8.0Gb in `.nii`. It's a 181x218x181x288 f32 image, thus allocating 8 227 466 496 bytes and reading from a Gz source, here

    let mut raw_data = vec![0u8; nb_bytes_for_data(header)?];
    source.read_exact(&mut raw_data)?;

I tested and it doesn't seem to be a memory issue, in the sense that it does reach the `read_exact` line, but then it's stuck for, err, long enough that I kill the job. 7zip decodes it in ~1m40s, nifti-rs reads the non-gz version in ~10s. For the gz version, it allocates ~3750Mb, then run indefinitely (max we waited was 1 hour) while always using one process, so it's doing something.

We will probably work with HCP image in the future so we might want to contribute a solution to this problem. I'm not sure how to solve this though! Do you think a chunk version would work? Something like:

    out = image of right dimension
    buffer = vec![0; 1024]
    while not eof
        read chuck
        reinterpret to input type
        cast to requested type
        linear_transform
        assign to out  at right place.
    return out

It might slow down the reading of "normal"/smaller images, but we can probably create a different code path for "big" images. What do you think?

Enet4:
Coincidently, one of the ideas that I've had at the back of my head for a while was a "lazy" NIfTI volume implementation, which would not pull all of the volume from the file to memory. It would be backed by some sort of paged caching mechanism, thus restricting memory usage while preventing an excessive number of file reads. Combined with adaptor methods for retrieving volume slices, I believe that such a volume would solve that particular problem.

On the other hand, it's in fact weird that the program can allocate enough memory for the volume, but fail to read it afterwards. Something else might be at play here, so I would like to look into this as well.

fmorency:
I've been bitten by this bug while processing huge multishell DWI datasets. The workaround of using uncompressed NIFTI files works at the expense of disk space.

Enet4:
I also agree that this is an important (and fairly challenging) matter. So far, I've thought of two non-exclusive ways to overcome the problem:

[...]

PR #52, opened 2019-06-27
Streamed Volumes

Enet4:
This gathers a bunch of additions and changes to the crate so as to provide slice-by-slice reading of volumes. (#44)

A summary of all changes below:

- Split `NiftiVolume` trait to have a separate random access API: `RandomAccessNiftiVolume`
- Added `Dim` and `Idx` type definitions for containing a NIfTI-compliant volume shape and index to volume, provides a nicer abstraction than working with the 8-element array directly.
- New `volume::streamed` module contains a volume implementation which reads the voxel data slice by slice.
- Deprecated `InMemNiftiVolume::new_from_stream` and all other `from_stream` associated functions in favor of `from_reader`, which is more idiomatic and does not conflict with the futures concept of stream
- Made `GenericNiftiObject`. `InMemoryNiftiObject` and `StreamedNiftiObject` are now aliases to this type.
- Added information to `NiftiError::IncompatibleLength`


src/object.rs
```diff
@@ -171,56 +202,323 @@ impl InMemNiftiObject {
...
+    /// # fn run() -> Result<()> {
+    /// let obj = StreamedNiftiObject::from_file("minimal.nii.gz")?;
+    /// 
+    /// let volume = obj.into_volume();
+    /// for slice in volume {
+    ///     // manipulate slice here
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
```
nilgoyette: Is this supposed to be indented like a normal function would be?

src/object.rs
```diff
@@ -171,56 +202,323 @@ impl InMemNiftiObject {
...
+        let len = header.vox_offset as usize;
+        let len = if len < 352 { 0 } else { len - 352 };
+        let extender = Extender::from_reader(&mut source)?;
+        
+        let ext = {
+            let source = ByteOrdered::runtime(&mut source, header.endianness);
+            ExtensionSequence::from_reader(extender, source, len)?
+        };
+
+        let volume = InMemNiftiVolume::from_reader(source, &header)?;
```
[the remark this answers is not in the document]
nilgoyette: Now that I read more, I understand that you're probably aware of this code copy. `from_file` and `from_file_2` being identical. Probably others too. Do you have a plan to fix this?

src/util.rs
```diff
@@ -64,14 +67,40 @@ where
...
+        return Err(NiftiError::InconsistentDim(i as u8, raw_dim[i]));
+    }
+    Ok(o)
+}
+
+pub fn validate_dimensionality(raw_dim: &[u16; 8]) -> Result<usize> {
+    if raw_dim[0] == 0 || raw_dim[0] > 7 {
+        return Err(NiftiError::InconsistentDim(0, raw_dim[0]));
+    }
+    Ok(usize::from(raw_dim[0]))
```
nilgoyette: Why not use the good old `as usize`? Isn't it identical?

src/volume/streamed.rs
```diff
@@ -0,0 +1,449 @@
...
+            dim: [3, 2, 3, 2, 0, 0, 0, 0],
+            datatype: NiftiType::Uint16 as i16,
+            scl_slope: 1.,
+            scl_inter: 0.,
+            endianness: Endianness::Little,
+            ..NiftiHeader::default()
+        };
+
+        let mut volume =
+            StreamedNiftiVolume::from_reader_rank(&volume_data[..], &header, 1).unwrap();
```
nilgoyette: I really don't understand why you need to create a slice out of `volume_data`, which is already a slice. Ah, hum, it works if you write

    let volume_data: &[u8] = &[

I thought that the right type would be automatically known by the compiler.

tests/volume.rs
```diff
@@ -182,6 +183,33 @@ mod ndarray_volumes {
...
+        }
+        assert!(slices[4].iter().any(|v| *v != 0.));
+        assert!(slices[5].iter().any(|v| *v != 0.));
+
+        assert_ulps_eq!(slices[0][[5, 0]], 0.0);
+        assert_ulps_eq!(slices[0][[0, 5]], 0.0);
+        assert_ulps_eq!(slices[5][[0, 0]], 0.0);
+        assert_ulps_eq!(slices[4][[5, 0]], 0.4_f32 as f64);
+        assert_ulps_eq!(slices[5][[0, 8]], 0.8_f32 as f64);
+        assert_ulps_eq!(slices[5][[5, 5]], 1.0_f32 as f64);
```
nilgoyette: Err, what?

src/volume/lazy.rs
```diff
@@ -0,0 +1,48 @@
...
+    source: F,
+}
+
+impl<F> LazyNiftiVolume<F>
+where
+    F: Read,
+    F: Seek,
+{
+
+}
```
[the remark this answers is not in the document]
Enet4: Oops, I committed this file by mistake. Off it goes.

src/object.rs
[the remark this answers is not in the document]
Enet4: You're right, a lot of the code is replicated and I am likely to refactor this part before merging. What might make things a bit more complicated is that in-memory and streamed volumes have a different set of options (`slice_rank`), which means that a hypothetical volume-from-reader interface would need to keep this into account.

src/object.rs
Enet4: Since the outer function constructs are invisible in the doctests, it is preferable not to indent the body. What could be done now is take advantage of `?` in doctests (since 1.34) and discard the wrapper boilerplate entirely. I've done so in a few other places, one might as well do the same for new functions.

src/util.rs
Enet4: Not exactly. `usize::from` does not allow a conversion from a wider type, whereas a cast using `as` can truncate to narrower types. Using `as` in cases where the conversion should be lossless [triggers a Clippy lint](https://rust-lang.github.io/rust-clippy/current/index.html#cast_lossless).

src/volume/streamed.rs
Enet4: This is one of the cases where the compiler doesn't perform deref coercion. The input for `from_reader_rank` is something that implements `Read`, but references to arrays don't implement that and the compiler won't turn it into a slice automatically.

tests/volume.rs
Enet4: Well, to facilitate the testing process, I collected all of the slices and did the equivalent value testing routine as the one for the in-memory volume. `slices[i][[j, k]]` would be fetching the value in position [j, k, i] (since data is in column major order).

Enet4:
Thanks for the initial review, @nilgoyette. A few more changes were pushed just now, some of which reduce redundancy. The subtle differences in each kind of volume make it pretty hard to merge common routines even further.

I haven't benchmarked this yet, but it would be nice to see some measurements on large volumes. The real benefit comes from using much less memory, at the cost of potentially a bit of extra maintenance overhead (plus more allocations if the idiomatic iterator is used instead of the streaming iterator).

src/object.rs
```diff
@@ -64,48 +95,269 @@ impl InMemNiftiObject {
...
+        R: Read,
+        V: FromSource<R>,
+    {
+        // fetch extensions
+        let len = header.vox_offset as usize;
+        let len = if len < 352 { 0 } else { len - 352 };
+
+        let ext = {
+            let source = ByteOrdered::runtime(&mut source, header.endianness);
+            ExtensionSequence::from_reader::<_, _>(extender, source, len)?
```
nilgoyette: Remove `::<_, _>`

src/object.rs
```diff
@@ -64,48 +95,269 @@ impl InMemNiftiObject {
...
+
+        let file = BufReader::new(File::open(&hdr_path)?);
+        if gz {
+            Self::from_file_pair_2(GzDecoder::new(file), vol_path, Default::default())
+        } else {
+            Self::from_file_pair_2(file, vol_path, Default::default())
+        }
+    }
+    /*
+        fn from_file_pair_2<S, Q>(mut hdr_stream: S, vol_path: Q) -> Result<InMemNiftiObject>
```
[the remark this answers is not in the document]
Enet4: Oopsie daisy.

src/object.rs
```diff
@@ -64,48 +95,269 @@ impl InMemNiftiObject {
...
+        let len = header.vox_offset as usize;
+        let len = if len < 352 { 0 } else { len - 352 };
+        let extender = Extender::from_reader(&mut source)?;
+
+        let ext = {
+            let source = ByteOrdered::runtime(&mut source, header.endianness);
+            ExtensionSequence::from_reader(extender, source, len)?
+        };
+
+        let volume = V::from_reader(source, &header, Default::default())?;
```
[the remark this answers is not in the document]
Enet4: Yes, great catch!

src/object.rs
Enet4: The turbo fish will be hooked.

PR #53, opened 2019-07-08
Migrate to Rust 2018 Edition

Enet4:
The minimum required compiler version is already way beyond 1.31 with the latest updates, so we might as well benefit from the latest edition.


nilgoyette:
I have 4 warnings with a `cargo build`, in `inmem.rs` and `object.rs`.

Enet4:
Oops. Thanks for the heads up!

nilgoyette:
LGTM. Thank you.

PR #54, opened 2019-07-10
Use Either type to reduce object impl redundancy

Enet4:
This makes a significant improvement in the maintainability of the `object` module.
By including the `either` crate, we obtain a sum type for data which may be conditionally GZip encoded, that still implements `Read`.

Minor caveat: this now uses the `Either` type for reading the volume's data, even for in-memory volumes. This should hopefully not introduce a noticeable overhead.


nilgoyette:
Nice cleanup, thank you. lgtm.

PR #55, opened 2019-07-11
Remove nifti header builder

Enet4:
This removes the `NiftiHeaderBuilder` and the dependency `derive_builder`.
Reasons:
- It was not very useful, considering that we can build a default instance and work from there.
- Error handling isn't ideal, the error type is `String` which does not match how errors emerge from the rest of the crate, and I'm not sure if this would ever return an error, considering that all fields have a default value.
- The builder could leave the false impression that fields are validated when using it, which was never really the case.
- I could not just deprecate it because the implementation is behind a `derive` macro.

Hopefully there's no one relying heavily on this builder.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ repository = "Enet4/nifti-rs"
 
 [dependencies]
 byteordered = "0.4.0"
-derive_builder = "0.7.0"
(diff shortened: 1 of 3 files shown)


nilgoyette:
I don't know about other programmers, but I confirm that we never used it. LGTM.

Issue #41, opened 2019-01-29
Rethink some of the header's default attribute values


Enet4:
This was discussed in #40, but I'm filing an issue so that this concern is not forgotten.

> - `pixdim` should be [1.0; 8] by default.
> - `srow_x` and other should have an "identity" default.
> - `sform_code` should be 1 by default IF it's really the normal way to encode the transformation. Maybe  `qform_code` is the way, I don't know,

I believe we're still undecided on what should be the defaults for `sform_code` and `qform_code`.

nilgoyette:
My colleague @jchoude tells me that we should do:

- reading
  - Normal case: if **either** `sform_code` or `qform_code` is present (not 0), apply the requested one.
  - Special case: both are present. Standard says nothing about this case but most tools use `sform`
- writing: calculate and write the 2 transformations with `sform_code = 1` and `qform_code = 0`. We will have to wait for #22 to do it properly though.

This doesn't directly answer the question, but as I understand it, the default Header should be

- "identity" in both transformations
- `sform_code = 1` 
- `qform_code = 0`

jchoude:
Just a minor correction, @nilgoyette , the `qform_code` should also be `1`. `0` means that it's not recorded.

Enet4:
I believe this has been resolved in #56. As usual, this can be brought up again if something is worth rethinking.

PR #56, opened 2019-07-11
Change defaults in header

Enet4:
As discussed in #41, some fields have more reasonable default values than the ones we currently have. Are there any other fields that we should look into? (`qform_code`/`sform_code`?)

- `pixdim`: `[1.;8]`
- `srow_x`: `[1, 0, 0, 0]`
- `srow_y`: `[0, 1, 0, 0]`
- `srow_z`: `[0, 0, 1, 0]`


src/header.rs
```diff
@@ -207,9 +207,9 @@ impl Default for NiftiHeader {
             quatern_y: 0.,
             quatern_z: 0.,
 
-            srow_x: [0.; 4],
-            srow_y: [0.; 4],
-            srow_z: [0.; 4],
+            srow_x: [1., 0., 0., 0.,],
+            srow_y: [0., 1., 0., 0.,],
+            srow_z: [0., 0., 1., 0.,],
```
nilgoyette: I asked my colleague again, just to be sure that he hasn't changed his mind in the last 6 months. In his [educated] opinion, we should set identity to sform AND qform, and set both codes to 1. Of course, the standard is not clear on everything, so setting both to the right values is a safe bet.

If you don't want to set both, then at least set `sform_code` to 1, otherwise it tells the reader to ignore the sform matrix. At least one of the `code` should be `!= 0`.

tests/header.rs
```diff
@@ -13,6 +13,9 @@ fn minimal_hdr() {
         datatype: 2,
         bitpix: 8,
         pixdim: [0., 3., 3., 3., 0., 0., 0., 0.],
+        srow_x: [0.; 4],
+        srow_y: [0.; 4],
+        srow_z: [0.; 4],
         vox_offset: 0.,
         scl_slope: 0.,
         scl_inter: 0.,
```
nilgoyette: You might want to create a `get_test_header` or something for those 15-lines header creations. At least the first three are identical.

Enet4: Indeed, I will create helper functions for building the headers for "minimal" test files.

Enet4:
I have rebased this branch and included the suggested changes. It seems that the writer does behave differently based on whether the header states "sform_code = 2, qform_code = 1" or "sform_code = 2, qform_code = 0".

tests/writer.rs
```diff
@@ -275,6 +275,7 @@ mod tests {
             datatype: NiftiType::Rgb24 as i16,
             pixdim: [1.0; 8],
             sform_code: 2,
+            qform_code: 0,
             srow_x: [1.0, 0.0, 0.0, 0.0],
             srow_y: [0.0, 1.0, 0.0, 0.0],
             srow_z: [0.0, 0.0, 1.0, 0.0],
```
nilgoyette: You can remove `pixdim` and `srow_#` on the three tests. And maybe add a `get_rgb_header` function to avoid repeating those 7 lines.

tests/writer.rs
```diff
@@ -310,6 +311,7 @@ mod tests {
             datatype: NiftiType::Rgb24 as i16,
             pixdim: [1.0; 8],
             sform_code: 2,
+            qform_code: 0,
```
nilgoyette: > It seems that the writer does behave differently

You made me worried with this. It's only because we compare the image bytes by bytes. Changing the right byte like we do in lines 287-293 or changing it for real in the RGB image on disk should work too. However, keeping `qform_code: 0,` is simpler. As you wish.

Enet4: Thank you for clearing this out. Please let me know if you think this test should be made more transparent of this matter.

nilgoyette: You forgot to remove `pixdim` ;)

And, yes, it's probably a good idea to add a comment in `rgb_header_gt` explaining why `qform_code: 0,` otherwise we will forget soon enough. Or if you have a hex editor, just change the byte from 0 to 1 (I can also do it in another branch, you don't have to do everything).

Enet4: Oops. Hopefully this is it.

nilgoyette:
LGTM. Good job.

Enet4:
I think it's high time we had a release. I'll work on it later today.

PR #59, opened 2019-09-25
ndarray 0.13 + build_dim_array

### LINKED ISSUE #58: ndarray 0.13
I tried updating nifti-rs to ndarray 0.13 because we also want to update our enterprise project, but there's a problem. Now that "approx" is a feature flag of ndarray, we need a feature that doesn't exist: [optional `dev-dependencies`](https://github.com/rust-lang/cargo/issues/1596). Without this feature, the "approx" crate will always be included, even in normal builds. I understand that we do not want that, but I don't see what we're supposed to do about it. Nobody's working in cargo issue 1596, so we shouldn't wait. Thoughts?

nilgoyette:
Fix #58 

- Update to ndarray 0.13 and required fixes (from_iter, all_close and approx). Version 0.13 is now required for nifiti-rs.
- As discussed, I moved the approx crate from "dependencies" to "dev-dependencies".
- There some autoformat in this PR. I added `#[rustfmt::skip]` 3 times to avoid matrix formatting. I think formatting in Rust is great but not for matrices! I don't like using skip, but I consider it essential for matrices.
- The `Create build_dim_array function` commit doesn't fit in this PR, sorry. I'm not a fan of creating PR for tiny changes like that, but maybe you prefer it cleanly separated. Just tell me and I won't do it again :)

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -550,26 +550,6 @@ impl NiftiHeader {
     }
 }
 
-/// Build the `dim` field of the `NiftiHeader` from a ndarray shape.
-///
-/// # Example
-///
-/// ```
-/// use ndarray::Array3;
-/// use nifti::header::build_dim_array;
-/// let im = Array3::<f32>::zeros((3, 4, 5));
-/// assert_eq!(build_dim_array(im.shape()), [3, 3, 4, 5, 1, 1, 1, 1]);
-/// ```
-#[cfg(feature = "ndarray_volumes")]
-pub fn build_dim_array(data_shape: &[usize]) -> [u16; 8] {
-    let mut dim = [1; 8];
-    dim[0] = data_shape.len() as u16;
-    for (i, s) in data_shape.iter().enumerate() {
-        dim[i + 1] = *s as u16;
-    }
-    dim
-}
-
 fn parse_header_1<S>(input: S) -> Result<NiftiHeader>
 where
     S: Read,
--- a/src/header.rs
+++ b/src/header.rs
@@ -386,23 +386,12 @@ impl NiftiHeader {
         T: RealField,
         f32: SubsetOf<T>,
     {
+        #[rustfmt::skip]
         let affine = Matrix4::new(
-            self.srow_x[0],
-            self.srow_x[1],
-            self.srow_x[2],
-            self.srow_x[3],
-            self.srow_y[0],
-            self.srow_y[1],
-            self.srow_y[2],
-            self.srow_y[3],
-            self.srow_z[0],
-            self.srow_z[1],
-            self.srow_z[2],
-            self.srow_z[3],
-            0.0,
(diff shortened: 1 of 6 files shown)


src/header.rs
```diff
@@ -547,6 +550,26 @@ impl NiftiHeader {
...
+/// # Example
+///
+/// ```
+/// use ndarray::Array3;
+/// use nifti::header::build_dim_array;
+/// let im = Array3::<f32>::zeros((3, 4, 5));
+/// assert_eq!(build_dim_array(im.shape()), [3, 3, 4, 5, 1, 1, 1, 1]);
+/// ```
+#[cfg(feature = "ndarray_volumes")]
+pub fn build_dim_array(data_shape: &[usize]) -> [u16; 8] {
```
Enet4: If this is to be a public method, we should highlight the fact that it panics: add a `# Panics` section in the documentation and make it fail faster by checking the shape's length sooner.

```suggestion
pub fn build_dim_array(data_shape: &[usize]) -> [u16; 8] {
    if data_shape > 7 {
        panic!("cannot make NIfTI-1 shape of {} dimensions (maximum is 7)", data_shape.len());
    }
```

src/writer.rs
```diff
@@ -182,7 +176,7 @@ where
     };
 
     let mut header = NiftiHeader {
-        dim,
+        dim: build_dim_array(data.shape()),
```
Enet4: At this level however, I question whether it should panic or error. It might be uncertain how much control the consumer has of the provided array. What do you think? If it should error instead, we can just add a some kind of test like the one below and make it pass.

```rust
let path = get_temporary_path(path);
let data = ArrayD::zeros(&[1; 8]);
assert!(write_nifti(path, data, None).is_err());
```

nilgoyette: Good questions. I don't really want this function to be public, but the outside tests also use it, so this might be an indication that this function can be useful to others.

I think the best is to be consistant. We already have `validate_dim` and `validate_dimensionality`, which return an error. I'll move the new function there, rename it/them and make it return an error.

Enet4: Oh by the way, we already have [`Shape::from_slice`](https://github.com/Enet4/nifti-rs/blob/master/src/volume/shape.rs#L140), which takes a slice of `u16`. With a few num traits, we should be able to make it generic for any slice of unsigned numbers.

nilgoyette:
As there's no specific PR or issues that we're currently working on, would it be possible to get a release with ndarray 0.13? Maybe updating `num-derive` before, if you do a release.

`nifti-rs` is the only missing lib for ndarray 0.13 in our project and we're kind of excited about it :)

Enet4:
Yes, it's a nice time for another release, but we might as well merge #61 and think a bit about #60 first. :+1:

PR #64, opened 2020-02-05
Dimension 1 bug

### LINKED ISSUE #63: Writing 4D images with a single volume
This snippet currently produces a mangled image, as if the bytes are written in the wrong order.

    let mut data = Array4::zeros((5, 5, 4, 3));
    data.slice_mut(s![.., 2, 2, 0]).fill(1.0);
    data.slice_mut(s![.., 3, 2, 0]).fill(1.1);
    write_nifti("/tmp/test.nii.gz", &data.select(Axis(3), &[0]), None).unwrap();

Changing the indices to `&[0, 1]` works well. This tells me that writing a 4D image with a single volume is broken. I had no success fixing it yet but I'm working on it.

nilgoyette:
I'm not entirely happy with this fix because I don't understand why it works :| It make sense that writing the data bytes of a (1, 5, 5, 5) image is the same as writing them for a (5, 5, 5) image. The first dimension adds nothing interesting that should concern the `write_data` function. However, it looks like iterating on a dim=1 change the ordering. When I skip the dim=1, it works as intended.

Anyway, I'm still pushing this branch because I have other things to do, but if you have a better understanding of the problem than me, I'm interested!

Fix #63

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -290,7 +290,13 @@ where
     W: Write,
     E: Endian + Copy,
 {
-    let mut iter = data.axis_iter(Axis(0));
+    // We always write the image by iterating on the first axis, thus
+    //   3D (depth, height, width) => per axial slice
+    //   4D (time, depth, height, width) => per volume
+    // However, there's a problem when dim == (1, ...). We must iterate by the first non-one length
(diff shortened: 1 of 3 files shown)


nilgoyette:
I requested you as reviewer because I don't even know if you saw this PR.

PR #70, opened 2020-06-11
Add ReaderOptions and ReaderStreamedOptions

nilgoyette:
- Make InMemNiftiObject's methods privates
- Make StreamedNiftiObject's methods privates
- Remove the deprecated new_from_stream method
- Apply one fix on header when requested
- Fix all tests

Sorry, there was some automatic formatting when I saved (like in lib.rs). The other concerns I have at the moment are:
- One can still read the complete image by creating a File, reading the header, then reading the volume. Should I block that? By making some methods `pub crate` in `header.rs` and `volume/inmen.rs`?
- I don't want to block it too much because we want a way to read only the header without reading the complete image. We're already using this in our codebase. It's practical and much faster.
- I separated the features in `ReaderOptions` and `ReaderStreamedOptions` to avoid repetitive names `read_file_streamed/_rank/_pair/_pair_rank` (I can at least remove the 'streamed' in all names. But it does create some code copy. If we ever add some general methods, like `read_header`, the code will be exactly the same. So, I'm not sure it was the right decision.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -18,7 +18,7 @@ use std::path::Path;
 pub use crate::util::{GzDecodedFile, MaybeGzDecodedFile};
 
 /// Options and flags which can be used to configure how a NIfTI image is read.
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct ReaderOptions {
     /// Whether to automatically fix value in the header
     fix_header: bool,
@@ -94,7 +94,7 @@ impl ReaderOptions {
 }
 
 /// Options and flags which can be used to configure how a NIfTI image is read and iterated.
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct ReaderStreamedOptions {
     /// Whether to automatically fix value in the header
     fix_header: bool,
--- a/src/object.rs
+++ b/src/object.rs
@@ -304,7 +304,10 @@ impl InMemNiftiObject {
     /// let obj = InMemNiftiObject::from_file("minimal.nii.gz")?;
     /// # Ok::<(), nifti::NiftiError>(())
     /// ```
-    #[deprecated(note = "use `read_file` from `ReaderOptions` instead")]
+    #[deprecated(
+        since = "0.10.0",
+        note = "use `read_file` from `ReaderOptions` instead"
+    )]
     pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
         let gz = is_gz_file(&path);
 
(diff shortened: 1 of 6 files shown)


src/object.rs
```diff
@@ -94,6 +304,7 @@ impl InMemNiftiObject {
     /// let obj = InMemNiftiObject::from_file("minimal.nii.gz")?;
     /// # Ok::<(), nifti::NiftiError>(())
     /// ```
+    #[deprecated(note = "use `read_file` from `ReaderOptions` instead")]
```
Enet4, line 307:
```suggestion
    #[deprecated(since = "0.10.0", note = "use `read_file` from `ReaderOptions` instead")]
```

src/object.rs
```diff
@@ -108,6 +319,7 @@ impl InMemNiftiObject {
     /// Retrieve a NIFTI object as separate header and volume files.
     /// This method is useful when file names are not conventional for a
     /// NIFTI file pair.
+    #[deprecated(note = "use `read_file_pair` from `ReaderOptions` instead")]
```
Enet4, line 322:
```suggestion
    #[deprecated(since = "0.10.0", note = "use `read_file_pair` from `ReaderOptions` instead")]
```

src/object.rs
```diff
@@ -156,6 +362,7 @@ impl StreamedNiftiObject<MaybeGzDecodedFile> {
     /// }
     /// # Ok::<(), nifti::NiftiError>(())
     /// ```
+    #[deprecated(note = "use `read_file` from `ReaderStreamedOptions` instead")]
```
Enet4, line 365:
```suggestion
    #[deprecated(since = "0.10.0", note = "use `read_file` from `ReaderStreamedOptions` instead")]
```

src/object.rs
```diff
@@ -167,6 +374,7 @@ impl StreamedNiftiObject<MaybeGzDecodedFile> {
     /// If the file only contains the header, this method will
     /// look for the corresponding file with the extension ".img",
     /// or ".img.gz" if the former wasn't found.
+    #[deprecated(note = "use `read_file_rank` from `ReaderStreamedOptions` instead")]
```
Enet4, line 377:
```suggestion
    #[deprecated(since = "0.10.0", note = "use `read_file_rank` from `ReaderStreamedOptions` instead")]
```

src/object.rs
```diff
@@ -190,6 +398,7 @@ impl StreamedNiftiObject<MaybeGzDecodedFile> {
     /// }
     /// # Ok::<(), nifti::NiftiError>(())
     /// ```
+    #[deprecated(note = "use `read_file_pair` from `ReaderStreamedOptions` instead")]
```
Enet4, line 401:
```suggestion
    #[deprecated(since = "0.10.0", note = "use `read_file_pair` from `ReaderStreamedOptions` instead")]
```

src/object.rs
```diff
@@ -203,6 +412,7 @@ impl StreamedNiftiObject<MaybeGzDecodedFile> {
     /// streamed volume reading, using `slice_rank` as the dimensionality of
     /// each slice. This method is useful when file names are not conventional
     /// for a NIfTI file pair.
+    #[deprecated(note = "use `read_file_pair_rank` from `ReaderStreamedOptions` instead")]
```
Enet4, line 415:
```suggestion
    #[deprecated(since = "0.10.0", note = "use `read_file_pair_rank` from `ReaderStreamedOptions` instead")]
```

PR #72, opened 2020-07-04
Remove stuff that was deprecated since 0.8.0

Enet4:
Should be safe to take them out at this point,
considering that they were hard-deprecated in 0.8 and 0.9.


nilgoyette:
Lgtm.

PR #74, opened 2020-07-21
Add the 7 optional FSL intents

### LINKED ISSUE #73: FSL intent
I was wondering if we should add the remaining intent codes in nifti-rs, which stops at

    #define NIFTI_INTENT_SHAPE        2005

According to the standard, there are 7 others intents (`support is OPTIONAL for conforming implementations.`)

    #define NIFTI_INTENT_FSL_FNIRT_DISPLACEMENT_FIELD               2006
    #define NIFTI_INTENT_FSL_CUBIC_SPLINE_COEFFICIENTS              2007
    #define NIFTI_INTENT_FSL_DCT_COEFFICIENTS                       2008
    #define NIFTI_INTENT_FSL_QUADRATIC_SPLINE_COEFFICIENTS          2009
    #define NIFTI_INTENT_FSL_TOPUP_CUBIC_SPLINE_COEFFICIENTS        2016
    #define NIFTI_INTENT_FSL_TOPUP_QUADRATIC_SPLINE_COEFFICIENTS    2017
    #define NIFTI_INTENT_FSL_TOPUP_FIELD                            2018

I ask because of course we deal with some of those images. It made my nifti header printer crash. I don't **need** to modify nifti-rs to repair it, I'll fix it wathever happens, but I think we should add those intents because they do appear in the standard.

nilgoyette:
- Some automatic formatting in the `use`.
- I simply added the missing intents in the enum. I don't think there's anything else to do.
- There's no documentation in the standard for those fields, so I simply rewrote the name in a human readable way.
- I wasn't sure if `FSLFNIRTDisplacementField` or `FslFnirtDisplacementField` was better...?

Fix #73

### DIFF
--- a/src/typedef.rs
+++ b/src/typedef.rs
@@ -416,20 +416,20 @@ pub enum Intent {
     Shape = 2005,
 
     /// FSL FNIRT Displacement field.
-    FSLFNIRTDisplacementField = 2006,
+    FslFnirtDisplacementField = 2006,
     /// FSL Cubic spline coefficients.
-    FSLCubicSplineCoefficients = 2007,
+    FslCubicSplineCoefficients = 2007,
     /// FSL Discrete cosine transform coefficients.
-    FSLDCTCoefficients = 2008,
+    FslDctCoefficients = 2008,
     /// FSL Quadratic spline coefficients.
-    FSLQuadraticSplineCoefficients = 2009,
+    FslQuadraticSplineCoefficients = 2009,
 
     /// FSL Topup cubic spline coefficients.
-    FSLTopupCubicSplineCoefficients = 2016,
+    FslTopupCubicSplineCoefficients = 2016,
     /// FSL Topup quadratic spline coefficients.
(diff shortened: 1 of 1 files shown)


Enet4:
> - Some automatic formatting in the use.
> - I simply added the missing intents in the enum. I don't think there's anything else to do.
> - There's no documentation in the standard for those fields, so I simply rewrote the name in a human readable way.

All good!

> - I wasn't sure if `FSLFNIRTDisplacementField` or `FslFnirtDisplacementField` was better...?

The [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#c-case) state that the latter is the conventional form. On the other hand, I am guilty of not always favouring conventions over familiarity in other projects, although I am not entirely sure how important this factor is for an intent in nifti. I can give you an opportunity to rename these variants, should you wish these variants to comply with the guidelines, otherwise we're good to merge.

nilgoyette:
Thank you for the information, I'll update them to respect the guidelines.

PR #75, opened 2020-08-23
Writer options

### LINKED ISSUE #49: Automatic extension
One of my users complained that ".nii" wasn't appended to the filename he provided to my program. I understand that there's no obvious answer to this question, but I think that a lib called `nifti-rs` should probably save ".nii" files by default instead of directly using the filename provided without question. What do you think about that?

nilgoyette:
Here's a first draft of `WriterOptions`. I don't see anything particularly horrible or insulting to a programmer :) Maybe

- `header_file: bool` is not a super name. Does nifti have a name for the distinct files (nii vs hdr+img)?
- I feel that the path update is "unclean". Maybe we should calculate the final path only in the `write` functions?
- I wanted to add an option to set a reference path, in case the user doesn't want to load the header himself and simply give us the path to the header, but we would lose the [faster] `reference: Option<&'a NiftiHeader>`. If we don't care about copying this struct and we love this feature, I can do it. It would also remove the lifetime, which some people may like :)
- Do you have other ideas of useful parameters?

> This would also be an opportunity to ground one common API for writing, regardless of their data type (RGB) or how they are represented (ndarray or a NiftiVolume type).

I still don't understand how I can write a single function for "normal" **and** rgb image. I see that they now share some common parameters, but we still have the problem with the traits. `[u8; 3]` must respect `DataElement` and `TriviallyTransmutable`. Because of this, I kept the `write_nifti` and `write_rgb_nifti` functions (now transformed to methods).

Fix #49 and maybe some other stuff.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -17,196 +17,259 @@ use crate::{
     DataElement, NiftiHeader, NiftiType, Result,
 };
 
-/// Write a nifti file (.nii or .nii.gz).
-///
-/// If a `reference` is given, it will be used to fill most of the header's fields. Otherwise, a
-/// default `NiftiHeader` will be built and written. In all cases, the `dim`, `datatype` and
-/// `bitpix` fields will depend only on `data`, not on the header. This means that the `datatype`
-/// defined in `reference` will be ignored. Because of this, `scl_slope` will be set to 1.0 and
-/// `scl_inter` to 0.0.
-pub fn write_nifti<P, A, S, D>(
-    header_path: P,
-    data: &ArrayBase<S, D>,
-    reference: Option<&NiftiHeader>,
-) -> Result<()>
-where
-    P: AsRef<Path>,
-    S: Data<Elem = A>,
-    A: DataElement,
-    A: TriviallyTransmutable,
-    D: Dimension + RemoveAxis,
-{
-    let compression_level = Compression::fast();
-    let is_gz = is_gz_file(&header_path);
-    let (header, data_path) =
-        prepare_header_and_paths(&header_path, data, reference, A::DATA_TYPE)?;
+/// Options and flags which can be used to configure how a NIfTI image is written.
+#[derive(Debug, Clone, PartialEq)]
+pub struct WriterOptions<'a> {
+    /// Where to write the output image (header and/or data).
+    header_path: PathBuf,
+    /// If given, it will be used to fill most of the header's fields. Otherwise, a default
+    /// `NiftiHeader` will be built and written. In all cases, the `dim`, `datatype` and `bitpix`
+    /// fields will depend only on `data` parameter passed to the `write` function, not on this
+    /// field. This means that the `datatype` defined in `reference` will be ignored. Because of
+    /// this, `scl_slope` will be set to 1.0 and `scl_inter` to 0.0.
+    reference: Option<&'a NiftiHeader>,
+    /// Whether to write the header and data in distinct files. (nii vs hdr+img)
+    header_file: bool,
+    /// Whether to compress the output to gz. Is guessed from the `header_path`, but can be
+    /// overriden with the `compress` method, which will update the `header_path` to keep coherency.
+    compress: bool,
+    /// Compression level to use when writing to the gz file.
+    compression_level: Compression,
+}
 
-    // Need the transpose for fortran ordering used in nifti file format.
-    let data = data.t();
-
-    let header_file = File::create(header_path)?;
-    if header.vox_offset > 0.0 {
-        if is_gz {
-            let mut writer = ByteOrdered::runtime(
-                GzEncoder::new(header_file, compression_level),
-                header.endianness,
-            );
-            write_header(writer.as_mut(), &header)?;
-            write_data::<_, A, _, _, _, _>(writer.as_mut(), data)?;
-            let _ = writer.into_inner().finish()?;
-        } else {
-            let mut writer = ByteOrdered::runtime(BufWriter::new(header_file), header.endianness);
-            write_header(writer.as_mut(), &header)?;
-            write_data::<_, A, _, _, _, _>(writer, data)?;
+impl<'a> WriterOptions<'a> {
+    /// Creates a blank new set of options ready for configuration.
+    pub fn new<P>(header_path: P) -> WriterOptions<'a>
+    where
+        P: AsRef<Path>,
+    {
+        let mut header_path = header_path.as_ref().to_owned();
+        if header_path.extension().is_none() {
+            let _ = header_path.set_extension("nii");
         }
-    } else {
-        let data_file = File::create(&data_path)?;
-        if is_gz {
-            let mut writer = ByteOrdered::runtime(
-                GzEncoder::new(header_file, compression_level),
-                header.endianness,
-            );
-            write_header(writer.as_mut(), &header)?;
-            let _ = writer.into_inner().finish()?;
-
-            let mut writer = ByteOrdered::runtime(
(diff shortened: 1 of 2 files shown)


src/writer.rs
```diff
@@ -17,196 +17,259 @@ use crate::{
...
+    /// Where to write the output image (header and/or data).
+    header_path: PathBuf,
+    /// If given, it will be used to fill most of the header's fields. Otherwise, a default
+    /// `NiftiHeader` will be built and written. In all cases, the `dim`, `datatype` and `bitpix`
+    /// fields will depend only on `data` parameter passed to the `write` function, not on this
+    /// field. This means that the `datatype` defined in `reference` will be ignored. Because of
+    /// this, `scl_slope` will be set to 1.0 and `scl_inter` to 0.0.
+    reference: Option<&'a NiftiHeader>,
+    /// Whether to write the header and data in distinct files. (nii vs hdr+img)
+    header_file: bool,
```
Enet4, line 32: So far the nomenclature has narrowed to either "NIfTI file pair" or "header and volume files". Both should be clear, given the context.

On the other hand, the name `header_file` can be a bit confusing, considering that we also have `header_path`. A better name IMO is `write_header_file`.

src/writer.rs
```diff
@@ -17,196 +17,259 @@ use crate::{
...
-            let _ = writer.into_inner().finish()?;
-        } else {
-            let mut writer = ByteOrdered::runtime(BufWriter::new(header_file), header.endianness);
-            write_header(writer.as_mut(), &header)?;
-            write_data::<_, A, _, _, _, _>(writer, data)?;
+/// Options and flags which can be used to configure how a NIfTI image is written.
+#[derive(Debug, Clone, PartialEq)]
+pub struct WriterOptions<'a> {
+    /// Where to write the output image (header and/or data).
+    header_path: PathBuf,
```
Enet4, line 24: As I understood this one, the field can either represent the path to the header file or the path to the full NIfTI file, right? Maybe it is better to call it `path` and explain in the documentation that it represents one or the other based on the `write_header_file` field.

I am not sure of whether there is interest in being able to manually specify both the path to the .hdr file _and_ the .img file. Should we later on find this in the wild, this field could become of some new enum type `WriteDestination` for increased flexibility. But I don't believe that we need this at the moment.

src/writer.rs
```diff
@@ -329,3 +392,39 @@ where
...
+    #[test]
+    fn test_writer_path() {
+        let w = WriterOptions::new("~/image");
+        assert_eq!(w.header_path, PathBuf::from("~/image.nii"));
+
+        for (p, nii, niigz) in &[
+            ("~/image", "~/image.nii", "~/image.nii.gz"),
+            ("~/image.nii", "~/image.nii", "~/image.nii.gz"),
+            ("~/image.nii.gz", "~/image.nii", "~/image.nii.gz"),
+            ("~/image.hdr", "~/image.hdr", "~/image.hdr.gz"),
```
Enet4, line 409: Do these tests demonstrate that enabling compression on an options with a header file path `"~/image.hdr"` would convert it to `"~/image.hdr.gz"`, and therefore compress the header? I could be mistaken, but is there interest in compressing both header and volume files? The reader only supports this for flexibility, but my impression is that header files are more often left uncompressed even if the respective .img is compressed.

Enet4 (commented):
Thank you for working on this. Here is my first review, I might look into more of it and extend the review later.

> I feel that the path update is "unclean". Maybe we should calculate the final path only in the write functions?

That might indeed be better, so as to reduce the number of unnecessary mutations.
We can constrain the other setter methods to simple validation (most of which is already done by the type system).

> I wanted to add an option to set a reference path, in case the user doesn't want to load the header himself and simply give us the path to the header, but we would lose the [faster] reference: Option<&'a NiftiHeader>. If we don't care about copying this struct and we love this feature, I can do it. It would also remove the lifetime, which some people may like :)

I am a bit intrigued to see how that would look like. Maybe an additional enum layer would be the best way of representing the three possible combinations (no header, let the program infer everything; reference to in-memory header; path to header file).

nilgoyette:
Sorry, I totally forgot about this review! I'm on a 5 weeks vacation (only 3 left) because I'm a dad now :) As you can guess, I'm not much into programming right now. I'll probably return to this review when I'm back to work in october.

Enet4:
No worries, take your time! And congratulations!

src/writer.rs
nilgoyette: I removed the support for writing compressed headers because, as you wrote, it's mostly useless. Asking for

    WriterOptions::new("brain.nii.gz").write_header_file(true).write_nifti(&data).unwrap();
    or
    WriterOptions::new("brain.nii").write_header_file(true).compress(true).write_nifti(&data).unwrap();

would write `brain.nii.gz` and `brain.hdr`. However, what if the user asks

    WriterOptions::new("brain.hdr.gz").write_header_file(true).write_nifti(&data).unwrap();

With my current version, we write `brain.nii.gz` and `brain.hdr`... I'm pretty sure that the user wants`brain.hdr.gz`. Of course, this broke `write_hdr_standard` which write a hdr and hdr.gz file, then tries to read the same path.

What's your opinion on this? Should we support writing compressed headers only when the user specifically asks for it?

nilgoyette:
I tried to code the "reference from file" idea.

    #[derive(Debug, Clone, PartialEq)]
    enum Reference<'a> {
        None,
        FromHeader(&'a NiftiHeader),
        FromFile(PathBuf), //  I don't think it's possible to use &'a Path here
    }
    ...
    pub struct WriterOptions<'a> {
        ...
        reference: Reference<'a>,
    ...
    impl<'a> WriterOptions<'a> {
        ...
        pub fn reference_header(mut self, reference: &'a NiftiHeader) -> Self {
            self.reference = Reference::FromHeader(reference);
        ...
        pub fn reference_file<P>(mut self, path: P) -> Self
        where
            P: AsRef<Path>,
        {
            self.reference = Reference::FromFile(path.as_ref().to_path_buf());
        ...
        fn prepare_header_and_paths<T, D>(...) {
            let reference = match &self.reference {
                Reference::FromHeader(h) => h.clone().to_owned(),
                Reference::FromFile(path) => NiftiHeader::from_file(path)?,
                Reference::None => {
                    let mut header = NiftiHeader::default();
                    header.sform_code = 2;
                    header
                }
            };

Do you know how to make this cleaner? Or do we simply forget about this idea?

src/writer.rs
Enet4: Granted, if the user wrote "whatever.hdr.gz", then they are clearly expecting the header to be compressed. I think we should honor that, even if this means adding another property to our writer options.

In any other case (such as stating "whatever.nii[.gz]"), I say we always write the header uncompressed.

Enet4:
```rust
#[derive(Debug, Clone, PartialEq)]
enum Reference<'a> {
    None,
    FromHeader(&'a NiftiHeader),
    FromFile(PathBuf), //  I don't think it's possible to use &'a Path here
}
```

I don't see the problem in using `&'a Path`, so long as the `P` type parameter is also constrained to the lifetime `'a`.

Other than this, the logic for turning this `Reference` type into a dedicated `NiftiHeader` can be put in a self-consuming method. I'm also torn between the names `Reference` and `HeaderReference`, as the former feels a bit too generic.

nilgoyette:
Oh, I always write `path: P` without thinking, but `path: &'a P` was necessary in this case, otherwise `path` was dropped while still borrowed. Thank you.

Enet4 (approved):
All right, these are good to go! Thank you for your patience and for keeping up with this.

PR #76, opened 2020-11-05
Update libs

nilgoyette:
I had to remove the `default-features = false` for nalgebra because there was sudenly a lot of missing stuff. I tried adding the right features instead, to no avail.

As for `quick_error`, I replaced `cause` with `source` and everything's still ok, but I don't actually know what I'm doing ;)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,12 +29,11 @@ num-derive = "0.3"
 num-traits = "0.2"
 quick-error = "1.2"
 safe-transmute = "0.11"
-either = "1.5"
+either = "1.6"
 
 [dependencies.nalgebra]
-default-features = false
 optional = true
-version = "0.21"
+version = "0.23"
 
 [dependencies.ndarray]
 optional = true
@@ -44,7 +43,7 @@ features = ["approx"]
 [dependencies.simba]
 default-features = false
 optional = true
-version = "0.1"
+version = "0.3"
 
 [dev-dependencies]
 pretty_assertions = "0.6"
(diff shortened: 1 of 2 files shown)


Enet4:
Hah, you got to this first! ;) Granted, it's easier to find out that some dependencies are outdated now that [deps.rs](https://github.com/deps-rs/deps.rs) has been revived.

> As for `quick_error`, I replaced `cause` with `source` and everything's still ok, but I don't actually know what I'm doing ;)

That is accurate. This version of quick-error seems to be removing its use of deprecated portions of the standard Error API. We were already not using  `description` anymore, and the only thing left was to start using `source` instead of `cause` (also [deprecated in `Error`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.cause)).

With nothing else to question here, I can merge right away. Thanks!

nilgoyette:
Sorry to bother you with that, but if you don't have any other merge that you want to do before a while, we would appreciate a release so we can use nalgebra 0.23 in our codebase.

nilgoyette:
Wait, what the hell, I forgot to update `approx` :| Sigh

nilgoyette:
Nevermind, we're better not touch `approx` until the next release of `ndarray`. Sorry for spamming you like that!

nilgoyette:
Hum, you didn't answer my last request with a yes or no so I'll ask again in case you didn't see it.

> Sorry to bother you with that, but if you don't have any other merge that you want to do before a while, we would appreciate a release so we can use nalgebra 0.23 in our codebase.

Enet4:
Right, sorry about that. :x These changes have just been released in 0.11.0.

PR #77, opened 2021-01-08
Update libs

nilgoyette:
I didn't think this through the last I asked you for a release! I couldn't even use it ^^ because I also needed ndarray 0.14 otherwise I had some weird dependency prpoblem with approx. Anyway, now everything will be updated, yay.


Enet4:
Yay indeed! Thanks.

PR #79, opened 2021-01-09
Add ndarray_volumes as a default Cargo feature

Enet4:
Truth be told, this crate is particularly useful when combined with ndarray,
there are way more use cases which require it than not.
So I say it's time to promote this feature as enabled by default.


nilgoyette:
I totally agree with this change.

PR #80, opened 2021-01-09
Clean documentation and readme

Enet4:
- Clean up documentation blocks
   - escape square brackets when they are not supposed to be links
   - code formatting for certain identifiers and mathematical expressions
   - fix an intra-link in the streamed module
- Update readme to use `ReaderOptions` instead of deprecated functions.
  This is pretty much the reason why I stick to very short examples in the readme file, if any.


nilgoyette:
`cargo doc` runs without problem on all features. +1

PR #81, opened 2021-01-10
Create Actions - rust.yml

Enet4:
Now that Travis CI has limited free computational use on public repositories, it's high time one had a transition to GitHub Actions.

While I don't believe that we can achieve an equivalent pipeline at this time, this should suffice for now, and more can be added later.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,56 @@
+on: [push, pull_request]
+
+name: Continuous integration
+
+jobs:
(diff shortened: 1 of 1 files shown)


nilgoyette:
I always find those files totally unreadable. I'm not even gonna try to understand it. Don't wait for my opinion, merge when you want.

Enet4:
I know the feeling, I was hoping not to take more than half an hour to set this up, but at least it works now.

PR #82, opened 2021-01-10
Slap Travis CI

Enet4:
I suspect that Travis was using more computational credits than necessary. There were bugs in the file, yet Travis ran the jobs anyway. With #81 in, most of the jobs here don't need to stay in Travis anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,18 @@
 language: rust
 
-os: linux
-
-rust:
- - stable
- - beta
- - nightly
+env:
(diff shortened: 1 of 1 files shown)


Enet4:
Note to future readers: no violence is condoned from slapping a CI system, this is just an educational, very metaphorical, slap in the wrist.

nilgoyette:
Hey Eduardo, I hope I won't ask this again for a while, but can you please release version 0.12? Unless you're working on something awesome for nifti-rs that will be released soon!

Also, I don't mind violence against Travis ;)

PR #83, opened 2021-01-15
Remove traits for into_ndarray

nilgoyette:
It seems that `T: Mul<Output = T>,` and `T: Add<Output = T>,` are no longer used for the `into_ndarray` functions.


Enet4 (approved):
This seems reasonable indeed. `DataElement` already implies (albeit not formally) that it can be converted to values which support these arithmetical operations anyway.

PR #85, opened 2021-04-09
Update libs

nilgoyette:
With ndarray ecosystem (npy and stats) ready for 0.15, we might as well update too. I updated `pretty_assertions`, `nalgebra`+`simba` and `ndarray`. There was no specific fix or problem with those libs. And I don't think think we should update the code related to the first three. However, maybe we could update nifti-rs to actually use some of ndarray 0.15 new features, but I don't know this library precisely enough to know where to look. Maybe the best is to wait until we do have a real task and then use what's new.

I didn't update the minimum rust version because it doesn't exist in this library. It's now at Rust 1.49.

As always, if there's nothing in the near-future to change, my colleagues and me would appreciate a release with these changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ version = "0.24"
 
 [dependencies.ndarray]
 optional = true
-version = "0.14"
+version = "0.15"
 features = ["approx"]
 
 [dependencies.simba]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ optional = true
 version = "0.4"
 
 [dev-dependencies]
-pretty_assertions = "0.6"
+pretty_assertions = "0.7"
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
I feel that no changes are needed. Despite ndarray 0.15 comprising a substantial amount of changes, `nifti` already appears to work as it is and we might not even find any ways to improve this integration.

I'm not sure how complicated it usually is to migrate from ndarray 0.14 to 0.15, but it should hopefully not be a hindrance. There is nothing else "in store" at the moment, so users can stick to `nifti` 0.12.1 until they migrate the rest of the project anyway.

This is great, thank you! I will release version 0.13.0 right away.

Enet4:
It is done. I also took the liberty of updating a few stale things from the readme file.

PR #86, opened 2021-06-09
nalgebra 0.27.x & simba 0.5

nilgoyette:
Because of [this security advisory](https://rustsec.org/advisories/RUSTSEC-2021-0070), we thought that it's important to update asap nalgebra and simba to avoid any problems with a real bug or with a medical audit.

I didn't update `approx` to v0.5 because ndarray is still using v0.4.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ either = "1.6"
 
 [dependencies.nalgebra]
 optional = true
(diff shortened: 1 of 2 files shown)


Enet4:
Things seem to be in good order! This will imply a minor version bump, since `nalgebra` is part of the public API.

Thanks!

PR #88, opened 2021-06-10
Clippy fix + add Clippy to GitHub Actions

Enet4:
The cool thing about having an action for Clippy here is that detected problems are presented inline with the code changes.

The greatest change was taking the recommendation not to make exact comparisons in floating point types.
This was being done in header correction for `pixdim[0]`, to check whether it is either 1 or -1.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -21,9 +21,9 @@ use std::ops::Deref;
 use std::path::Path;
 
 /// Magic code for NIFTI-1 header files (extention ".hdr[.gz]").
-pub const MAGIC_CODE_NI1: &'static [u8; 4] = b"ni1\0";
+pub const MAGIC_CODE_NI1: &[u8; 4] = b"ni1\0";
 /// Magic code for full NIFTI-1 files (extention ".nii[.gz]").
-pub const MAGIC_CODE_NIP1: &'static [u8; 4] = b"n+1\0";
+pub const MAGIC_CODE_NIP1: &[u8; 4] = b"n+1\0";
 
 /// The NIFTI-1 header data type.
 /// All fields are public and named after the specification's header file.
@@ -237,7 +237,7 @@ impl NiftiHeader {
     /// Currently, only the following problems are fixed:
     /// - If `pixdim[0]` isn't equal to -1.0 or 1.0, it will be set to 1.0
     pub fn fix(&mut self) {
-        if self.pixdim[0].abs() != 1.0 {
+        if (self.pixdim[0].abs() - 1.).abs() > 1e-11 {
             self.pixdim[0] = 1.0;
         }
     }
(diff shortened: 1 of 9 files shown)


src/header.rs
```diff
@@ -237,7 +237,7 @@ impl NiftiHeader {
     /// Currently, only the following problems are fixed:
     /// - If `pixdim[0]` isn't equal to -1.0 or 1.0, it will be set to 1.0
     pub fn fix(&mut self) {
-        if self.pixdim[0].abs() != 1.0 {
+        if (self.pixdim[0].abs() - 1.).abs() > 1e-11 {
```
nilgoyette, line 240: I see that we use this comparison 2 times. It's nothing dramatic of course, but we might want to add a `is_pixdim_valid` method or something like that.

src/lib.rs
```diff
@@ -63,6 +63,7 @@
 //!
 #![deny(missing_debug_implementations)]
 #![warn(missing_docs, unused_extern_crates, trivial_casts, unused_results)]
+#![allow(clippy::unit_arg)]
```
nilgoyette, line 66: Do you remember why this was required? It seems strange.

src/writer.rs
```diff
@@ -27,12 +27,13 @@ enum HeaderReference<'a> {
 impl<'a> HeaderReference<'a> {
     fn to_header(&self) -> Result<NiftiHeader> {
         match self {
-            HeaderReference::FromHeader(h) => Ok(h.clone().to_owned()),
+            HeaderReference::FromHeader(h) => Ok((*h).to_owned()),
```
nilgoyette, line 30: I would have thought that `h.to_owned()` would work. Is it because `h` is a `&&NiftiHeader` (according to rust-analyzer)?

nilgoyette (commented):
Excellent idea to use Clippy!

src/lib.rs
Enet4: This is something I might look into deeper, but [`unit_arg`](https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg) was triggered by all those function calls with `Default::default()` as the last parameter, which resolved to a unit value. An example of such a call is [here](https://github.com/Enet4/nifti-rs/blob/master/src/object.rs#L61).

I currently find this to be a false positive, because the case for why this lint exists does not really apply. The point of `Default::default()` is to generalize over whichever associated type `Options` is, and the recommendation made by the lint does not even make any sense.

```none
warning: passing a unit value to a function
  --> src/object.rs:67:13
   |
67 |             InMemNiftiObject::from_file_impl(path, GzDecoder::new(file), Default::default())
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::unit_arg)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
   |
67 |             Default::default();
68 |             InMemNiftiObject::from_file_impl(path, GzDecoder::new(file), ())
   |
```

src/writer.rs
Enet4: Indeed, it is a reference to another reference, which makes `(*h)` necessary. Deref rules are sometimes a bit awkward. :)

src/header.rs
Enet4: I moved it to a separate function now.

nilgoyette (approved):
LGTM. Thank you for the changes and for answering my questions.

Enet4:
Thank you for the review! Looks like we're ready for a new release.

Enet4:
`nifti` 0.14.0 is in!

Meanwhile, regarding the `unit_arg` lint, I have just filed rust-lang/rust-clippy#7341.

PR #90, opened 2022-01-13
Support NIFTI-2, resolves #89

### LINKED ISSUE #89: support nifti-2 files
The [nifti-1](https://nifti.nimh.nih.gov/nifti-1) file format was finalized in 2007.  In 2011, the [nifti-2](https://nifti.nimh.nih.gov/nifti-2) format was created to support larger data sets.  Nifti-2 is now widely-used in the neuroimaging field, and furthermore, it serves as the basis for the CIFTI-2 format.

A decade later, nifti-rs supports nifti-1 but not nifti-2.  Attempting to read a nifti-2 file produces a `NiftiError::InvalidFormat`.  This is unfortunate, because nifti-rs is otherwise an excellent library with many advantages over its C/C++ and Python counterparts.

Is there interest in adding nifti-2 support to nifti-rs?  The [differences between nifti-1 and nifti-2](https://www.nitrc.org/forum/forum.php?thread_id=2148&forum_id=1941) are not great.  See also [Anderson Winkler's blog](https://brainder.org/2015/04/03/the-nifti-2-file-format/) and [nifti2.h](https://nifti.nimh.nih.gov/pub/dist/doc/nifti2.h).  There are [test data here](https://nifti.nimh.nih.gov/pub/dist/data/nifti2/).
- No new header fields are added
- The types of 30 existing header fields are enlarged (e.g., 32-bit float --> 64-bit double)
- 7 existing header fields that were unused are removed
- The header itself is (obviously) larger
- The header fields are stored in a different order
- The magic string is different

Supporting nifti-2 should therefore not be too difficult, but would require some deliberate changes to the API.  Since the two headers are so similar, it might make sense to create concrete `Nifti1Header` and `Nifti2Header` types, and make `NiftiHeader` an `enum` over both types with getter methods to extract the values of shared fields as the larger type.  This would let users who are _reading_ an image not have to care about the underlying version (nifti1 vs nifti2).  Users _writing_ an image would still have to manually populate the header fields for whichever version they intend to write out.

```
enum NiftiHeader {
    Nifti1Header(Nifti1Header),
    Nifti2Header(Nifti2Header),
}
impl NiftiHeader {
    pub fn slice_duration(&self) -> f64 {
        match *self {
            // promote nifti1 type to size of larger nifti2 type
            Nifti1Header(ref header) => header.slice_duration as f64,
            Nifti2Header(ref header) => header.slice_duration,
        }
    }
}
```

I welcome feedback about the API changes and, although my time is limited, I would be happy to submit incremental pull requests against a nifti-2 branch.

benkay86:
Adds support for the NIFTI-2 header format.  The PR is just enough to compile and run tests.  If approved, documentation and additional test cases will be needed in a subsequent PR.  These changes are _**not**_ backwards compatible.

* Types and methods outside the header have been changed to default to the larger NIFTI-2 types.  For example, `nifti::volume::shape::Dim` is now based off `u64` instead of `u16`.

* `NiftiHeader` is changed to an `enum` that contains either a `Nifti1Header` or `Nifti2Header` structure.  `NiftiHeader` implements getter/setter methods to modify the underlying header without having to worry about which version it is.  The underlying structure can also be accessed directly.

```rust
hdr = NiftiHeader::default();
hdr.slice_duration = 1.0; // no longer works
hdr.set_slice_duration(1.0); // OK
let slice_duration: f32 = hdr.get_slice_duration(); // no longer works
let slice_duration: f64 = hdr.get_slice_duration(); // OK
// Conversion to larger NIFTI-2 types always succeeds.
let mut hdr2: Nifti2Header = hdr.into_nifti2();
hdr2.slice_duration = 1.0; // OK
let slice_duration: f64 = hdr2.slice_duration; // OK
// Conversion to smaller NIFTI-1 types is fallible.
let mut hdr1: Nifti1Header = hdr.into_nifti1().unwrap();
hdr1.slice_duration = 1.0; // OK
let slice_duration: f32 = hdr1.slice_duration; // OK
```

Migration should otherwise be relatively painless.
* `NiftiHeader::from_file()` and `from_reader()` automatically detect the header version in the given file or input stream.
* `WriterOptions::write_nifti()` automatically writes out whichever header version is referenced by the `WriterOptions`.
* `NiftiHeader::into_nifti1()` and `into_nifti2()` conveniently convert between header versions.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -59,7 +59,7 @@ pub const MAGIC_CODE_NIP1: &[u8; 4] = b"n+1\0";
 #[derive(Debug, Clone, PartialEq)]
 pub struct NiftiHeader {
     /// Header size, must be 348
-    pub sizeof_hdr: i32,
+    pub sizeof_hdr: u32,
     /// Unused in NIFTI-1
     pub data_type: [u8; 10],
     /// Unused in NIFTI-1
@@ -73,6 +73,7 @@ pub struct NiftiHeader {
     /// MRI slice ordering
     pub dim_info: u8,
     /// Data array dimensions
+    /// Note in the NIFTI-1 specification this is actually [i16; 8].
     pub dim: [u16; 8],
     /// 1st intent parameter
     pub intent_p1: f32,
@@ -85,9 +86,11 @@ pub struct NiftiHeader {
     /// Defines the data type!
     pub datatype: i16,
     /// Number of bits per voxel
-    pub bitpix: i16,
+    /// Note in the NIFTI-1 specification this is actually i16.
+    pub bitpix: u16,
     /// First slice index
-    pub slice_start: i16,
+    /// Note in the NIFTI-1 specification this is actually i16.
+    pub slice_start: u16,
     /// Grid spacings
     pub pixdim: [f32; 8],
     /// Offset into .nii file to reach the volume
@@ -97,11 +100,12 @@ pub struct NiftiHeader {
     /// Data scaling: offset
     pub scl_inter: f32,
     /// Last slice index
-    pub slice_end: i16,
+    /// Note in the NIFTI-1 specification this is actually i16.
+    pub slice_end: u16,
     /// Slice timing order
-    pub slice_code: u8,
+    pub slice_code: i8,
     /// Units of `pixdim[1..4]`
-    pub xyzt_units: u8,
+    pub xyzt_units: i8,
     /// Max display intensity
     pub cal_max: f32,
     /// Min display intensity
@@ -116,7 +120,7 @@ pub struct NiftiHeader {
     pub glmin: i32,
 
     /// Any text you like
-    pub descrip: Vec<u8>,
+    pub descrip: [u8; 80],
     /// Auxiliary filename
     pub aux_file: [u8; 24],
     /// NIFTI_XFORM_* code
@@ -146,7 +150,7 @@ pub struct NiftiHeader {
     /// 'name' or meaning of data
     pub intent_name: [u8; 16],
 
-    /// Magic code. Must be `b"ni1\0"` or `b"ni+\0"`
+    /// Magic code. Must be `b"ni1\0"` or `b"n+1\0"`
     pub magic: [u8; 4],
 
     /// Original data Endianness
@@ -185,7 +189,7 @@ impl Default for NiftiHeader {
             glmax: 0,
             glmin: 0,
 
-            descrip: vec![0; 80],
+            descrip: [0; 80],
             aux_file: [0; 24],
             qform_code: 1,
             sform_code: 1,
@@ -608,17 +612,17 @@ where
     h.intent_p3 = input.read_f32()?;
     h.intent_code = input.read_i16()?;
     h.datatype = input.read_i16()?;
-    h.bitpix = input.read_i16()?;
-    h.slice_start = input.read_i16()?;
+    h.bitpix = input.read_i16()? as u16;
+    h.slice_start = input.read_i16()? as u16;
     for v in &mut h.pixdim {
         *v = input.read_f32()?;
     }
     h.vox_offset = input.read_f32()?;
     h.scl_slope = input.read_f32()?;
     h.scl_inter = input.read_f32()?;
-    h.slice_end = input.read_i16()?;
-    h.slice_code = input.read_u8()?;
-    h.xyzt_units = input.read_u8()?;
+    h.slice_end = input.read_i16()? as u16;
+    h.slice_code = input.read_i8()?;
+    h.xyzt_units = input.read_i8()?;
     h.cal_max = input.read_f32()?;
     h.cal_min = input.read_f32()?;
     h.slice_duration = input.read_f32()?;
@@ -627,7 +631,7 @@ where
     h.glmin = input.read_i32()?;
 
     // descrip is 80-elem vec already
-    input.read_exact(h.descrip.as_mut_slice())?;
+    input.read_exact(&mut h.descrip)?;
     input.read_exact(&mut h.aux_file)?;
     h.qform_code = input.read_i16()?;
     h.sform_code = input.read_i16()?;
--- a/src/header.rs
+++ b/src/header.rs
@@ -1658,16 +1658,32 @@ impl NiftiHeader {
     }
 }
 
-fn parse_header_1<S>(input: S) -> Result<NiftiHeader>
+// Take any object that implements the `byteordered::Endian` trait and return a
+// dynamic `byteordered::Endianness` object.
+fn endianness<E: Endian>(e: E) -> Endianness {
+    if e.is_native() {
+        Endianness::native()
+    } else {
+        Endianness::native().to_opposite()
+    }
+}
+
+// Private function to parse a NIfTI-1 header with the given header size.
+// The `ByteOrdered` input stream must already be set to the correct endianness,
+// and it must be located at the first field after sizeof_hdr.
+fn parse_nifti1_header<S, E>(mut input: ByteOrdered<S, E>, sizeof_hdr: u32) -> Result<Nifti1Header>
 where
     S: Read,
+    E: Endian + Copy,
 {
-    let mut h = NiftiHeader::default();
+    // Initialize default header.
+    let mut h: Nifti1Header = Nifti1Header::default();
+    // Set size of header to given (already read in) size.
+    h.sizeof_hdr = sizeof_hdr;
+    // Set endianness from ByteOrdered input stream.
+    h.endianness = endianness(input.endianness());
 
-    // try the system's native endianness first
-    let mut input = ByteOrdered::native(input);
-
-    h.sizeof_hdr = input.read_i32()?;
+    // Read remaining header fields.
     input.read_exact(&mut h.data_type)?;
     input.read_exact(&mut h.db_name)?;
     h.extents = input.read_i32()?;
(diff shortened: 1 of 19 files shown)


src/header.rs
```diff
@@ -657,3 +1766,74 @@ where
...
+    }
+    h.slice_code = input.read_i32()?;
+    h.xyzt_units = input.read_i32()?;
+    h.intent_code = input.read_i32()?;
+    input.read_exact(&mut h.intent_name)?;
+    h.dim_info = input.read_u8()?;
+
+    // All done, return header with populated fields.
+    Ok(h)
+}
```
nilgoyette, line 1839: Do you use an automatic formatter?

src/object.rs
```diff
@@ -500,11 +502,11 @@ impl<V> GenericNiftiObject<V> {
         V: FromSource<R>,
     {
         // fetch extensions
-        let len = header.vox_offset as usize;
-        let len = if len < 352 { 0 } else { len - 352 };
+        let len: usize = header.get_vox_offset()?.try_into()?;
+        let len = if len == 0 { 0 } else { len - TryInto::<usize>::try_into(header.get_sizeof_hdr())? }; // TODO!
```
nilgoyette, line 506: Can you explain more what's left TODO?

tests/writer.rs
```diff
@@ -256,23 +258,10 @@ mod tests {
         assert_eq!(new_data, data);
     }
 
-    #[test]
-    fn write_descrip_panic() {
```
nilgoyette, line 260: Is this not tested anymore?

nilgoyette (commented):
I can't review anything thoroughly now, but here are some questions and comments.

Btw, good job. This is a lot of work. Thank you.

src/header.rs
```diff
@@ -202,167 +1398,69 @@ impl Default for NiftiHeader {
...
     }
+}
 
-    /// Check whether `pixdim[0]` is either -1 or 1.
-    #[inline]
-    fn is_pixdim_0_valid(&self) -> bool {
-        (self.pixdim[0].abs() - 1.).abs() < 1e-11
+impl Nifti2Header {
+    /// Place this `Nifti2Header` into a version-agnostic [`NiftiHeader`] enum.
+    pub fn into_nifti(self) -> NiftiHeader {
```
Enet4, line 1462: It would be idiomatic to also have implementations of `From<Nifti1Header>` and `From<Nifti2Header>` for `NiftiHeader`. This way, a conversion would be as simple as `header.into()`.

src/header.rs
```diff
@@ -379,11 +1477,11 @@ impl NiftiHeader {
     pub fn affine<T>(&self) -> Matrix4<T>
     where
         T: RealField,
-        f32: SubsetOf<T>,
+        f64: SubsetOf<T>,
```
Enet4, line 1480: This seems to suggest that all affine transformations will from now on use `f64` for calculations, regardless of the nature of the original data. Volume data value transformations also seem to have changed accordingly. Not sure whether this breaks any expectations to existing consumers (@nilgoyette?).

src/volume/element.rs
```diff
@@ -42,17 +42,17 @@ pub trait LinearTransform<T: 'static + Copy> {
 /// affine transformation, then converted back to the original type. Ideal for
 /// small, low precision types such as `u8` and `i16`.
 #[derive(Debug)]
-pub struct LinearTransformViaF32;
+pub struct LinearTransformViaF32; // TODO remove this for NIFTI-2?
```
Enet4, line 45: Were these changes merely a consequence of the wider nifti header? Perhaps by letting the API consumer tap into nifti-1, they could keep linear transformations in f32.

tests/writer.rs
Enet4: It is no longer possible because `descrip` was turned back into an 80-byte array. It was kind of what I originally had in mind, but large arrays were not as nice to work with in Rust. That changed in recent versions, so unless we witness a significant performance regression, we can save the description in a `[u8; 80]` here instead of `Vec<u8>`.

Enet4 (changes requested):
Thank you for this major contribution, @benkay86! I went through the code and made a quick review. What I feel are the two major concerns are:

- getter naming convention (as already mentioned, they should not use `get_`)
- linear transformations for volume conversion will use f64 instead of f32 in many cases, which I suspect could affect cases where the original data is all in f32. If all voxel values have to go through f64, this could translate to a bigger memory footprint and maybe less performance. It might help to try a stress test with f32 volumes and see if performance and memory usage changed significantly.

liningpan:
Hi, NIFTI2 support would be very helpful! 

Can we add additional intent codes from the HCP pipeline?

https://github.com/Washington-University/workbench/blob/master/src/FilesBase/nifti2.h#L33-L49
```cpp
const int32_t NIFTI_INTENT_CONNECTIVITY_UNKNOWN=3000;
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE=3001;
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE_TIME=3002;//CIFTI-1 name
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES=3002;//CIFTI-2 name
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED=3003;
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED_TIME=3004;//ditto
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SERIES=3004;
const int32_t NIFTI_INTENT_CONNECTIVITY_CONNECTIVITY_TRAJECTORY=3005;//ditto
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE_TRAJECTORY=3005;
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE_SCALARS=3006;
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE_LABELS=3007;
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SCALAR=3008;
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED_DENSE=3009;
const int32_t NIFTI_INTENT_CONNECTIVITY_DENSE_PARCELLATED=3010;
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SERIES=3011;
const int32_t NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SCALAR=3012;
```

However, several variables map to the same number which is not possible for rust enum.

benkay86:
Apologies, I had a baby and kind of lost track of this!  It looks like there is still broad interest in NIFTI-2 support, and aside from some formatting issues and the need for more test cases, the biggest obstacle is what to name getter methods.

@nilgoyette is right that Rust eschews getter methods in general, preferring the simplicity and explicitness of accessing a struct's field directly.  The current PR allows this by resolving the `NiftiHeader` enum to a `Nifti1Header` or `Nifti2Header` struct that can be manipulated directly.  In addition to this, convenient getter and setter methods which are agnostic to the underlying header version are provided for automatic conversion between field types.

In such cases where getter/setter methods are needed, the Rust convention is to name the setter for the field `foo` `set_foo()` and simply name the getter `foo()`.  However, there is a problem for nifti-rs because a few fields already have getter-like functions with the same name.  For example, `NiftiHeader::dim` is a `[u16;8]` and `NiftiHeader::dim()` returns a validated slice of this field.  *What should we name a getter that returns the raw field?*

Possibilities:
- Name raw getters `get_foo()`.  While this is verbose and goes against Rust convention, it is easily understood and avoids name conflicts with existing methods.
- Rename the existing methods something else like `foo_validated()` and name the raw getter `foo()`.  This is more idiomatic but breaks the API.  (Maybe not such a big deal since the PR contains other breaking changes in corner cases.)
- Don't provide a raw getter for fields that require some kind of validation or parsing.  Make the user cast the enum down to a `Nifti1Header` or `Nifti2Header` if raw access to the field is desired.

Thoughts?

liningpan:
I also tried to implement nifti-2 support myself. I took a more aggressive approach and cast nifti-1 and nifti-2 header to a common type derived from nifti-2. The common type also exposes Enum types directly instead of through a getter. I believe connectome workbench took a similar approach. However, this will complete nuke the current interface. 

I also took a few test cases from nibabel, for nifti-2 unit tests.

You can take a look my experimental branch here if you are interested. https://github.com/liningpan/nifti-rs/tree/nifti2

src/header.rs
```diff
@@ -24,42 +25,1125 @@ use std::path::Path;
 pub const MAGIC_CODE_NI1: &[u8; 4] = b"ni1\0";
 /// Magic code for full NIFTI-1 files (extention ".nii[.gz]").
 pub const MAGIC_CODE_NIP1: &[u8; 4] = b"n+1\0";
+/// Magic code for NIFTI-2 header files (extention ".hdr[.gz]").
+pub const MAGIC_CODE_NI2: &[u8; 8] = b"ni2\0\r\n\x1A\n";
+/// Magic code for full NIFTI-1 files (extention ".nii[.gz]").
+pub const MAGIC_CODE_NIP2: &[u8; 8] = b"n+2\0\r\n\x1A\n";
+
+/// Abstraction over NIFTI version 1 and 2 headers.  This is the high-level type
```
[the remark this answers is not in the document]
Enet4: I heard rumors that typewriters had issues which were resolved with the double spacing, and with that some people grew into the aesthetics of separating sentences like this. 🤔

Nowadays though, I practice a form of [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) or [semantic line breaks](https://sembr.org/), which recommends introducing a new line at the end of each sentence.

In any case, this isn't a critical enough concern to be a blocker.

Enet4:
> However, there is a problem for nifti-rs because a few fields already have getter-like functions with the same name. For example, `NiftiHeader::dim` is a `[u16;8]` and `NiftiHeader::dim()` returns a validated slice of this field. What should we name a getter that returns the raw field?

I don't take issue in introducing a breaking change to the API, especially for introducing such a big feature to the crate. `raw_dim` would make a decent method name for the slice as saved in the header (thus only available behind specific version headers).

Aside from this, I would also like to ensure that 1) it stays possible to work with a specific version explicitly, and that 2) calculations on NIfTI-1 can still use `f32` as common ground, as mentioned in [this comment](https://github.com/Enet4/nifti-rs/pull/90#discussion_r788069773).

tests/writer.rs
liningpan: smallvec? https://docs.rs/smallvec/latest/smallvec/

Issue #92, opened 2022-05-11
IncompatibleLength


nilgoyette:
Hi Eduardo,

nifti-rs panics at an unexpected place and I would like to have your opinion. We have the following method in inmem.rs:

    pub fn from_reader<R: Read>(source: R, header: &NiftiHeader) -> Result<Self> {
        // rather than pre-allocating for the full volume size, this will
        // pre-allocate up to a more reliable amount and feed the vector
        // sequentially, to prevent some trivial OOM attacks
        let nb_bytes = nb_bytes_for_data(header)?;
        let mut raw_data = Vec::with_capacity(nb_bytes.min(PREALLOC_MAX_SIZE));
        let nb_bytes_written =
            std::io::copy(&mut source.take(nb_bytes as u64), &mut raw_data)? as usize;

        if nb_bytes_written != nb_bytes {
            return Err(NiftiError::IncompatibleLength(nb_bytes_written, nb_bytes));
        }

        ...

Reading the comment and reading the code, I don't understand how it's supposed to work.
- The comment tells us it's ok to not read the whole data in a single read
- The line with `PREALLOC_MAX_SIZE` tells us the same thing
- Then there's the condition saying "Hey, if you didn't read the whole thing, explode"

What was the intention behind this code?

Enet4:
Hmm, the code was indeed an attempt to prevent panics from unruly input. The problem was not exactly in trying to read the whole data at once, but in trying to allocate enough memory at once, since that leads to a panic if the nifit file claims to have an unreasonably enormous volume that doesn't fit in memory.

With that said, it's a good thing that this was brought up. As of Rust 1.57.0, we can [`try_reserve`](https://doc.rust-lang.org/stable/alloc/vec/struct.Vec.html#method.try_reserve) our raw data vector instead, thus simplifying the code and potentially finding out what's wrong. Will throw in a PR soon.

nilgoyette:
Thank for working on this. It was simply a "No space left on device" problem. nifti-rs can hardly be blamed!

PR #93, opened 2022-05-11
Use try_reserve to allocate volume/extended data

Enet4:
- add error variants `ReserveVolume` and `ReserveExtended`
- Allocate the whole block at once
   - with `try_reserve_exact`, so it doesn't panic

#92 @nilgoyette


nilgoyette:
That does indeed make the intention clearer. Thank you. :+1:

Enet4:
Just ran a fuzzer on this and found no issues after a few minutes. Will merge this then.

PR #96, opened 2022-07-10
NIfTI-2 Support

liningpan:
This PR is based on #90 , and resolved most issues raised in the original PR, including getters / Into&From traits. The PR also resolved several merge conflicts with current master.

Regarding using f32 for nifti1 when every possible, I don't think it's quite possible with the current NiftiHeader design

```rust
pub enum NiftiHeader {
    /// The underlying header version is `NIFTI-1`.
    Nifti1Header(Nifti1Header),
    /// The underlying header version is `NIFTI-2`.
    Nifti2Header(Nifti2Header),
}
```

We might need to make NiftiHeader a trait with a floating number type defined separately for Nifti1 and Nifti2

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -23,7 +23,7 @@ where
 ///
 /// We get the translations from the center of the image (implied by `shape`).
 #[rustfmt::skip]
-pub(crate) fn shape_zoom_affine(shape: &[u16], spacing: &[f32]) -> Matrix4<f64> {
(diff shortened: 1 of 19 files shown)


nilgoyette:
Thank you @liningpan. I started reviewing but I stopped because I'm curious about @benkay86 opinion. Is this what you had in mind?

If I may ask, do you guys know each other? Why did you continue his work?

PR #97, opened 2022-11-16
Make the set_x_affine functions public

nilgoyette:
We realized while working on a terrible ANTs affine bug that those functions should be public, so that the programmer has more options than this method:

    pub fn set_affine<T>(&mut self, affine: &Matrix4<T>)
    where
        T: RealField,
        T: SubsetOf<f64>,
        T: ToPrimitive,
    {
        // Set affine into sform with default code.
        self.set_sform(affine, XForm::AlignedAnat);

        // Make qform 'unknown'.
        self.set_qform(affine, XForm::Unknown);
    }

This is our (Imeka) way of doing things, but it may not be everyone else's way.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -487,7 +487,7 @@ impl NiftiHeader {
     }
 
     /// Set affine transformation in 'sform' fields.
-    fn set_sform<T>(&mut self, affine: &Matrix4<T>, code: XForm)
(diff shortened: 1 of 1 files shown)


Enet4 (approved):
This is fine by me. The documentation on `set_sform` is a bit thin, but other than that, they are both documented and it should be relatively clear what is meant.

:+1:

PR #98, opened 2022-11-29
Make error type non-exhaustive

Enet4:
A common practice for most error types is to make them non-exhaustive, so that introducing more error variants does not break the API (as is already the case due to #93).


nilgoyette:
TIL. I didn't know about `non_exhaustive`. LGTM :)

PR #101, opened 2023-05-31
Feature/nifti extension writer (task 5 of issue #19)

twitzelbos:
Enable some basic writing of extension headers to NIFTI


src/writer.rs
```diff
@@ -253,12 +283,19 @@ impl<'a> WriterOptions<'a> {
         T: Data,
         D: Dimension,
     {
+        let mut vox_offset:f32 = 352.0;
+    
+        if let Some(extension_sequence) = self.extension_sequence.as_ref() {
+            if extension_sequence.bytes_on_disk() > 0 {
+                vox_offset += extension_sequence.bytes_on_disk() as f32;
+            }
```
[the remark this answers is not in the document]
twitzelbos: Done, thanks for the catch

twitzelbos:
> I never heard of nifti extension before, so I only judged the code quality. Looks good to me.
> 
> What would be nice is to read a tiny nifti generated by another tool, like NiBabel. We already have some tests doing that.

I'll be adding that shortly, extensions are used by some tools, but there is also a growing trend in the community to put json files into the NIFTI, which is why the extension::from_str() is probably, the one thing most people would want.

examples/gen_nifti/main.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::nifti::{writer::WriterOptions, Extender, Extension, ExtensionSequence};
+
+fn main() {
+    let mut args = env::args().skip(1);
+    let filename = args.next().expect("Path to NIFTI file is required");
+
+    // generate some test data 256x256 float32
+    let data = ndarray::Array3::<f32>::zeros((256, 256, 1));
+
+    let extension1 = Extension::new(8 + 4, 3, vec![0, 0, 0, 0]);
```
nilgoyette, line 16: @Enet4 I think our API could be greatly improved by removing the `esize: i32` argument.

    pub fn new(esize: i32, ecode: i32, edata: Vec<u8>) -> Self {
        if esize as usize != 8 + edata.len() panic!

This clearly indicates that we already know the answer and won't accept any other, so why ask it?

Enet4: Hmm true, the constructor is practically already encoding the extension size through `edata`, and we can reconstruct the `esize` field from that. A `new` function that does not ask for an `esize` would be easier to use and less error prone.

nilgoyette:
> How do you envision that test? I'm contributing a writer, this crate was already able to read extensions. I can provide a nibabel generated file, and prove that the one written by this crate is the same, when reading back?

Should it be **identical**, byte for byte? I guess it should, or maybe not because of endiannes. Then again, maybe that's not what we should test. Reading it and using `assert_eq! on the extensions` might be a better test.

twitzelbos:
Okay, I added a test. Showing binary compatibility wasn't easy because the default qform_code in this crate is 1, and in nibabel its 0, but I changed that in nibabel for the generation of the reference file.
'''
mini_array = np.zeros((8,8))
mini_ext_img = nib.Nifti1Image(mini_array, np.eye(4))
mini_ext_img.header['qform_code'] = 1
comment_extension = nib.nifti1.Nifti1Extension(6, b"Hello World!")
mini_ext_img.header.extensions.append(comment_extension)
nib.save(mini_ext_img, "minimal_extended_hdr.nii")
'''

twitzelbos:
Should be all good to go @nilgoyette @Enet4. I'd be glad to help on more things.

twitzelbos:
Sorry for the additional changes, decided to give it a bit more cleanup based on feedback by a coworker in our fork.

src/extension.rs
```diff
@@ -7,6 +7,29 @@
...
+    /// String Comment
+    NiftiEcodeComment = 6,
+    /// XCEDE extension in XML format
+    NiftiEcodeXCEDE = 8,
+    /// JimDimInfo
+    NiftiEcodeJimDimInfo = 10,
+    /// WorkflowFWDS
+    NiftiEcodeWorkflowFWDS = 12,
+    /// Freesurfer
+    NiftiEcodeFreesurfer = 14,
```
nilgoyette, line 31: What is your source of information for those constants? I usually use [this page](https://brainder.org/2012/09/23/the-nifti-file-format/) for everything related to nifti and I can only see 0, 2 and 4.

src/writer.rs
```diff
@@ -57,6 +57,9 @@ pub struct WriterOptions<'a> {
     /// The header file will only be compressed if the caller specifically asked for a path ending
     /// with "hdr.gz". Otherwise, only the volume will be compressed (if requested).
     force_header_compression: bool,
+
+    /// optional ExtensionSequence
```
nilgoyette, line 61: Can you please check all new doc `///` in this MR and ensure that they start with an upper case character?

In fact, I would prefer if you do it for all your new comments `//` but this is less important.

src/writer.rs
```diff
@@ -297,6 +320,34 @@ impl<'a> WriterOptions<'a> {
...
+    mut writer: ByteOrdered<W, E>,
+    extensions: Option<&ExtensionSequence>,
+) -> Result<()>
+where
+    W: Write,
+    E: Endian,
+{
+    let extensions = match extensions {
+        Some(extensions) => extensions,
+        None => { writer.write_u32(0)?; return Ok(())},
```
nilgoyette, line 333: Could you please configure your IDE to format on save? There's no automatic formatting in our cicd or elsewhere. rustfmt is here to avoid that kind of discussion :)

tests/writer.rs
```diff
@@ -407,4 +407,26 @@ mod tests {
             fs::read("resources/rgb/4D.nii").unwrap()
         );
     }
+
+    #[test]
+    fn write_extended_header() {
+        let data: ndarray::ArrayBase<ndarray::OwnedRepr<f64>, ndarray::Dim<[usize; 2]>> = Array2::zeros((8, 8));
```
nilgoyette, line 413: Why do you use this complex initialization? Isn't it the same as

    let data = Array2::<f64>::zeros((8, 8));
    or
    let data: Array2<f64> = Array2::zeros((8, 8));

src/extension.rs
twitzelbos: https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1_io.h
actually goes to 44 now.

tests/writer.rs
twitzelbos: Its what the compiler said, and being not super familar with ndarray I simply obliged. Will fix.

nilgoyette:
> This test did uncover that nibabel does pad extensions to lengths multiple of 16, as some documentations suggested

So, are we able to **read** files with non-padded extensions and extensions padded to 16? Is a file padded with multiple of 8 valid?

twitzelbos:
> > This test did uncover that nibabel does pad extensions to lengths multiple of 16, as some documentations suggested
> 
> So, are we able to **read** files with non-padded extensions and extensions padded to 16? Is a file padded with multiple of 8 valid?

I have not tried that, as we have written our extensions only with nibabel before. The NIFTI documentation does state that extensions must be padded to a multiple of 16 bytes (I think that is still because they want the data to start on a multiple of 16 (and vox_offset be a multiple of 16). Only few people use extensions, and both nibabel and this are now padding to 16 per extension, in agreement with what the Nifti 1.1 spec calls for, so we are good on this front. If it had been me designing this NIFTI format (there are many other things I'd have done differently), I would have just padded AFTER the extensions, so that the data starts on a multiple of 16, but it is what it is.

I could, in a separate PR explore the reading of the extensions more, and add them to the niftidump example, but its time to close this here, I think.

nilgoyette:
> multiple of 16, as some documentations suggested
> The NIFTI documentation does state that extensions must be padded to a multiple of 16

Sorry, my goal is not to make you work for nothing. I just wanted to clarify because "some documentations suggested" is unclear! Now that you wrote "NIFTI documentation does state", it's totally different! If this is what the NIFTI standard asks for, it's perfect!

I'll leave @Enet4 do his review and merge.

twitzelbos:
> 

Unfortunately the documentation is scattered and inconsistent in places. I should reach out to Box Cox to find out what the situation is now, and going forward, since he has retired.

src/header.rs
```diff
@@ -91,6 +91,7 @@ pub struct NiftiHeader {
     /// Grid spacings
     pub pixdim: [f32; 8],
     /// Offset into .nii file to reach the volume
+    /// Note: the highly unusual choice of f32 is intentional and due to trying to achieve header backwards compatibility to ANALYZE 7.5
```
Enet4, line 94: Adding a newline so that the small description does not get cluttered with the note.
```suggestion
    /// Offset into .nii file to reach the volume.
    ///
    /// Note: the highly unusual choice of f32 is intentional
    /// and due to trying to achieve header backwards compatibility to ANALYZE 7.5
```

Enet4 (commented):
Thank you for your contribution, @twitzelbos, it looks pretty solid (especially alongside #102)!

I will only suggest a few minor comment tweaks before merging.

twitzelbos:
> Thank you for your contribution, @twitzelbos, it looks pretty solid (especially alongside #102)!
> 
> I will only suggest a few minor comment tweaks before merging.

Thank you, I committed both of your suggestions!

Enet4 (approved):
Thank you!

PR #102, opened 2023-06-02
Feature/remove esize from extension interface

twitzelbos:
Removed esize from the Extension::new constructor
Added more ecodes


.github/workflows/rust.yml
```diff
@@ -2,9 +2,9 @@ name: Continuous integration
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
```
Enet4, line 5: The changes to this file may have been added by mistake.

Enet4 (commented):
I am likely to make a deeper review after #101 is merged.

.github/workflows/rust.yml
twitzelbos: Oops, yeah I renamed the master branch in our fork, because we don't do master branches no more. Let me see how to fix this.

Enet4 (commented):
One thing I overlooked in the previous pull request review is asking whether `NiftiEcode` should be marked as non-exhaustive, so that adding new variants does not break existing code. This should only be useful if the set of common extension codes tends to expand over time (maybe through custom integrations?).

I noticed that the type is currently not used anywhere, so it seems to be primarily a simple way to provide constants for known extension codes to developers writing new nifti files. If we go in this direction, we can also provide a getter at the `Extension` struct (in a similar fashion as getting the intent code from the header), and maybe add `#[non_exhaustive]`.

I added another minor recommendation inline.

twitzelbos:
> One thing I overlooked in the previous pull request review is asking whether `NiftiEcode` should be marked as non-exhaustive, so that adding new variants does not break existing code. This should only be useful if the set of common extension codes tends to expand over time (maybe through custom integrations?).
> 
> I noticed that the type is currently not used anywhere, so it seems to be primarily a simple way to provide constants for known extension codes to developers writing new nifti files. If we go in this direction, we can also provide a getter at the `Extension` struct (in a similar fashion as getting the intent code from the header), and maybe add `#[non_exhaustive]`.
> 
> I added another minor recommendation inline.

Yes, I created the enum for eventual future use, but haven't yet used it (in fact the extension codes I'm writing are not on this list). I can also remove the entire enum and save it for a future date, if that helps.

nilgoyette:
> I created the enum for eventual future use, but haven't yet used it (the extension codes I'm writing are not on this list).

If it's normal and common to create our own `NiftiEcode` , then I think an enum shouldn't be used. Unless we know that the majority of programmers use the standard ones. This is probably hard to know.

twitzelbos:
> 

Well, the extension codes in this enum are used by the respective software tools, so if you want to interact with data from those, its a nice thing to have. But many people also use their own private ones (i.e. we do), which of course wouldn't be part of an official package like this, but its nice to know which ones are officially claimed.

twitzelbos:
I have removed the enum type from this PR entirely, hopefully unblocking this. @Enet4 @nilgoyette

Enet4 (approved):
Things look in good order by me.

Regarding the extension codes, some food for thought: it might be best represented as a newtype encapsulating a `u32` or `i32`. This enables easy conversion back and forth and lets us declare associated constants for known codes, while maintaing its binary representation and keeping the value of the code itself mostly unrestrained.

Thank you very much for your commitment, @twitzelbos! (And thank you @nilgoyette for the reviews!)

PR #107, opened 2025-06-16
Update libs, run clippy

nilgoyette:
Hi @Enet4,

There wasn't much to release, but I think 2 years is enough for a new release :)

May I have your opinion on this warning?

    /// A trait that is both Read and Seek.
    pub trait ReadSeek: Read + Seek {}
              ^^^^^^^^
              trait `ReadSeek` is never used

I hesitate to remove it because it's a public trait. Anyone could use it... but Clippy should know better than to warn about an unused public trait? Do we simply tell clippy to ignore it?

Also, should we use a newer edition?

We're working on a release of our medical software and `nifti-rs` is the only crate left to update. When we merge this MR, could you please create a release?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,32 +18,32 @@ features = ["ndarray_volumes", "nalgebra_affine"]
 [dependencies]
 approx = "0.5"
 byteordered = "0.6"
-flate2 = "1.0"
-num-derive = "0.3"
+flate2 = "1.1"
+num-derive = "0.4"
 num-traits = "0.2"
 quick-error = "2.0"
-either = "1.6"
-num-complex = {version = "0.4.3", features=["bytemuck"]} 
-rgb = "0.8.36"
-bytemuck = {version = "1.13.1", features=["extern_crate_alloc"]}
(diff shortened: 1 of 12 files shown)


Enet4 (approved):
Yup, this sounds great! Thank you!

> I hesitate to remove it because it's a public trait. Anyone could use it... but Clippy should know better than to warn about an unused public trait? Do we simply tell clippy to ignore it?

Clippy warning about an unused public trait is only a riddle, until we see that the `util` module is private, so in the end it won't be visible to consumers.

In any case, considering that we about to deliver some updates with semver-breaking bumps in public dependencies, we could drop this trait even if it were public. :) I'll go ahead and push this to the PR.

Enet4:
> Also, should we use a newer edition?

For a moment I though this was already in 2021. Sure, I'll go ahead and bring this in and a few other goodies.

As for 2024, I believe that it is a bit too early for a library, and does not bring much benefit to impose such a big MSRV bump.

nilgoyette:
Oups, sorry, bad button!

